Issue #84, opened 2014-01-16
Automatically mount home directory in init script


bfirsh:
Shared folders is one of the things that is stopping me from switching myself and [Fig](http://orchardup.github.io/fig/) from docker-osx. It's essential for using Docker in development.

The way we have it set up for docker-osx is it mounts your home directory into the same place inside the VM. So, for example, `/Users/ben` on OS X is mounted at `/Users/ben` inside the VM. This works brilliantly for Docker volumes, because I can do run like `docker run -v .:/code ...` and it Just Works because `.` expands to `/Users/ben/...`.

I see there are plans to get NFS working in #64, but are there plans to do a similar automatic NFS mount for boot2docker? Or a better question perhaps – are there any objections to doing this? Because I'd quite like to do it. ;)

Issue #62, opened 2014-01-08
Can't mount a shared volume


schickling:
If I try to run a command in a docker container with a shared volume I get the following error: 

```
2014/01/08 20:10:22 Error: start: Invalid bind mount '/Users/johannes/Desktop/test:/var/www' : source doesn't exist
```

I'm using the new OSX binary (0.7.3). Please give me some feedback if this is related to the osx binary or this project.

PR #93, opened 2014-01-22
Added an hostonly interface with fixed ip address for guest and host

moul:
With this secondary interface we can now publish ports (`docker run ... -p 8000:80` for instance) without needs to configure new nat rules

The `docker` ip is `172.18.42.1`, the `host` ip is `172.18.42.254`

Since it's a secondary interface, it don't brake anything, you can still use `127.0.0.1:2022` and `127.0.0.1:4243`

TODO:
- [x] Add secondary interface
- [x] Create an auto-setup for /etc/hosts (like https://github.com/noplay/docker-osx/blob/master/docker-osx#L37-L61)
- [ ] Update the `DOCKER_HOST` in .bashrc ?
- [ ] Documentation

cc #85 (Give the VM a hostname in the init script)
cc https://github.com/steeve/boot2docker/pull/87 (NFS Client & Server, create a second host only network interface)
cc #84 (Automatically mount home directory in init script)
cc https://github.com/steeve/boot2docker/pull/73 (boot2docker / Forward all dynamic assigned ports)
cc #64 (nfs share / guest to host)
cc #104 (boot2docker / docker auto mapping port forwarding)
cc #62 (nfs volume)

### DIFF
--- a/boot2docker
+++ b/boot2docker
@@ -138,8 +138,34 @@ init() {
     $VBM storageattach $VM_NAME --storagectl "SATA" --port 0 --device 0 --type dvddrive --medium $BOOT2DOCKER_ISO
     $VBM storageattach $VM_NAME --storagectl "SATA" --port 1 --device 0 --type hdd --medium $VM_DISK
 
+    setup_etc_hosts
+
     log "Done."
     log "You can now type boot2docker up and wait for the VM to start."
+}
+
+setup_etc_hosts() {
+    if grep -q localdocker /etc/hosts; then
+        return
+    fi
+
+    line="$DOCKER_IP localdocker"
+
+    echo
+    echo "Adding localdocker alias to /etc/hosts (may need your password for sudo)."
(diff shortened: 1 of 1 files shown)


axelson:
For those curious about how it works in boot2docker-cli (which I installed with Boot2Docker.app) you need to expose the port as normal (using `-P` or `p 127.0.0.0:$HOST_PORT:$CONTAINER_PORT`), then you can access the port via the boot2docker vm ip, which is usually 192.168.59.103 (you should see the IP when you first run `boot2docker up`).

https://github.com/boot2docker/boot2docker-cli/issues/164

gusarov:
boot2docker-cli that been a solution to this now deprecated... I think issue is active again

tianon:
The solution today would be to do it via `docker-machine`, directly in VirtualBox, etc.  There's nothing in boot2docker itself we can change to help here, as the OS bits are functional.

Issue #357, opened 2014-05-14
boot2docker 0.8.1 on OS X: container DNS stuck on 8.8.8.8, 8.8.4.4


garthk:
My containers no longer reliably pick up my current DNS servers.

Check current settings:

```
$ cat /etc/resolv.conf
search internal1.example.com internal2.example.com internal3.example.com
nameserver 10.10.10.10
nameserver 10.11.10.11
```

Bounce `boot2docker`, as I've become accustomed to doing after VPN changes:

```
$ boot2docker down
[2014-05-14 16:06:39] Shutting down boot2docker-vm...

$ boot2docker up
[2014-05-14 16:06:44] Starting boot2docker-vm...
[2014-05-14 16:07:09] Started.

To connect the docker client to the Docker daemon, please set:
export DOCKER_HOST=tcp://localhost:4243
```

Check DNS settings on `boot2docker-vm`:

```
$ boot2docker ssh
docker@localhost's password: 
                        ##        .
                  ## ## ##       ==
               ## ## ## ##      ===
           /""""""""""""""""\___/ ===
      ~~~ {~~ ~~~~ ~~~ ~~~~ ~~ ~ /  ===- ~~~
           \______ o          __/
             \    \        __/
              \____\______/
 _                 _   ____     _            _
| |__   ___   ___ | |_|___ \ __| | ___   ___| | _____ _ __
| '_ \ / _ \ / _ \| __| __) / _` |/ _ \ / __| |/ / _ \ '__|
| |_) | (_) | (_) | |_ / __/ (_| | (_) | (__|   <  __/ |
|_.__/ \___/ \___/ \__|_____\__,_|\___/ \___|_|\_\___|_|
boot2docker: 0.8.1
             master : 17bb8c2 - Thu May  8 02:55:10 UTC 2014
docker@boot2docker:~$ cat /etc/resolv.conf
search internal1.example.com
nameserver 10.10.10.10
nameserver 10.11.10.11
```

The two `nameserver` entries are OK, but two of the three `search` domains vanished.

Ensure the `docker` process doesn't have DNS settings:

```
docker@boot2docker:~$ ps auxw|grep docker
  607 root     /usr/local/bin/docker -d -D -g /var/lib/docker -H unix:// -H tcp://0.0.0.0:4243
```

Good. So, the container should get the name servers we saw earlier, right? 

```
$ docker run --rm -t -i ubuntu:precise cat /etc/resolv.conf
nameserver 8.8.8.8
[...]

PR #411, opened 2014-06-28
#357 Wait for eth0 to be up before starting docker - borrow existing code

fayep:
Took code from tc-config which waits for dhcp to complete.  This version doesn't ping, but you could ping the DNS server (10.0.2.3) which I think only exists if you have nat dns host resolver enabled or 10.0.2.2 (does that always exist?)
This, tied with a change in boot2docker-cli which would create the VM with the nat dns host resolver (that I haven't yet codified) will make sure dns for containers is the same as the host dns.  I tested this manually :(

### DIFF
--- a/rootfs/rootfs/etc/rc.d/docker
+++ b/rootfs/rootfs/etc/rc.d/docker
@@ -1,3 +1,24 @@
 #!/bin/sh
 
+# dhcp.sh runs udhcpc async, so it exits before network is up
+wait4Server() {
+	/etc/init.d/dhcp.sh
+	SERVER=$1 ; NOPING=$2 ; CNT=0
+	if [ "$NOPING" == "1" ] ; then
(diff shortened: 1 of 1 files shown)


gesellix:
I needed to add a similar code due to `eth1` not being available before starting docker. see #824 for details. Is there any possibility to make a service start dependent on an interface?

Chekote:
I've been running into this issue continually, and this helped get me back on track. Thanks for saving my sanity!

gesellix:
would #960 also address the issue in this pull request?

tianon:
I think #960 probably helped somewhat here, and still believe our best solution would be to convince `dhcp.sh` to run synchronously.  

Regardless, I think this has run a little long in the tooth, and fresh approaches should be considered (if anything is necessary here).  For example, I don't think Docker for Windows or Docker for Mac suffer from this problem. :+1:

PR #497, opened 2014-08-22
Enable use of boot2docker.local hostname

md5:
This patch enables the use of the `boot2docker.local` hostname to access the local `boot2docker-vm` instance from the host.

Adding Avahi and its dependencies to the image to support this functionality adds about 2 megabytes to `boot2docker.iso`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,7 +56,8 @@ ENV TCZ_DEPS        iptables \
                     xz liblzma \
                     git expat2 libiconv libidn libgpg-error libgcrypt libssh2 \
                     nfs-utils tcp_wrappers portmap rpcbind libtirpc \
-                    curl ntpclient
+                    curl ntpclient \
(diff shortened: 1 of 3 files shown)


volkertb:
I managed to get it working by running a container instance of image enernoclabs/avahi, as follows:

    docker run --net host --name "avahi-daemon" -d enernoclabs/avahi

After entering the above command, the Docker machine (the VM running the Docker daemon) should become resolvable from the physical host machine (as well as other connected VMs) through its hostname, appended with the TLD ".local": 

    ping $DOCKER_MACHINE_NAME.local

mDNS/Zeroconf works out of the box on both Macs (through Bonjour) and most modern Linux distros (through Avahi). To add support for it on Windows, the Bonjour Print Services need to be installed, which Apple has made freely available at https://support.apple.com/kb/dl999.

Anyway, as you can all see, it's easy to set up in a Docker machine. Howeve, it would indeed be a great improvement if boot2docker supported this feature out of the box. Is the enernoclabs/avahi image really too big to integrate directly into boot2docker?

Issue #602, opened 2014-10-28
boot2docker does not use --registry-mirror option available with docker >= v1.3.0


ftom:
I am using boot2docker from OSX with version 1.3.0, the docker command line is not today exposing the new 1.3.0 option allowing to point to a local registry mirror (--registry-mirror=http://local-registry:5000) in order to allow docker images caching on local network. 
It would be really helpful for guys behind firewall or poor web access.

Checking docker --help options within boot2docker iso (through a boot2docker ssh), option is available regarding the embedded docker version.

Would it be possible to add it in the command line ?

Thank you !!

PR #604, opened 2014-10-28
boot2docker does not use --registry-mirror option available with docker ...

ftom:
...>= v1.3.0 #602

Added variable DOCKER_REG_MIRROR to detect and activate registry-mirror for Docker daemon
To be used via persistent partition :
echo "DOCKER_REG_MIRROR=http://<localregistry>" >> /var/lib/boot2docker/profile


tianon:
Closing given the age, the fact that b2d is in maintenance mode, and that what's requested is already possible via an existing environment variable.  Thanks!

PR #607, opened 2014-10-30
fix the automount issue (#480)

qxo:


tianon:
Sorry for the really long delay!

I think the comments/questions on https://github.com/boot2docker/boot2docker/pull/480 still apply, particularly:

> Do you have an easy way to test a Debian or Ubuntu live CD on the same setup to see if it has the same issue (thus telling us whether they've got some kind of solution for this already in place that we might be able to learn from or steal)?

I'm going to close for now, but if we can find one of the other large providers of live Linux distros doing something similar to this we can learn from and link back to, I'd love to implement something that's well-tested elsewhere (since we just don't have the testing resources or userbase of these other large distros).  Thanks!

PR #733, opened 2015-02-11
Enable openvswitch kernel module

nickethier:
With the rise of networking becoming more of a playground in docker it would be awesome if boot2docker supported the ovs kernel module. The total bytes after building the iso with this comes to 24117248.


tianon:
This is interesting, and obviously desired, but I'm not sure how I feel about it, especially since it takes us up roughly a full megabyte in size, and Docker itself doesn't officially have OVS support (yet).

nickethier:
Closing in favor of #740

Issue #339, opened 2014-05-07
Automount ... doesn't properly mount device with boot2docker-data label


Iristyle:
I've seen issue #216 

My setup is a little different.  I did the following (basically)
- Created a 10 GB VHD under Hyper-V
- Attach the boot2docker 0.8.0 ISO as a DVD drive
- Once the ISO has booted, I download the latest ISO inside of the VM, so that I can write to my 10GB VHD, which is at `/dev/sda` with `dd if=boot2docker.iso of=/dev/sda`
- I shutdown the VM, detach the ISO as a DVD drive, and boot from the VHD (which is using only the 24MB or so of the ISO)
- I use fdisk to create a new extended, then logical partition -- the extended is `/dev/sda2` and the logical is `/dev/sda5` -- it takes nearly all of the space, aside from the 24MB occupied in `/dev/sda1` by the image.
- I format the partition with `mkfs.ext4 -L boot2docker-data /dev/sda5` and everything looks good
- At this point, I reboot the image and expect that https://github.com/boot2docker/boot2docker/blob/master/rootfs/rootfs/etc/rc.d/automount will take over

Unfortunately, I end up with a `/mnt/sda5` but it's not properly symlinked, `/dev/sda5` is not mounted, and `/var/lib/docker` exists and is full of files.

I've gone through the process of manually executing bits of the script

```
BOOT2DOCKER_DATA=`blkid -o device -l -t LABEL=$LABEL`

...

PARTNAME=`echo "$BOOT2DOCKER_DATA" | sed 's/.*\///'`
```

properly sets `BOOT2DOCKER_DATA` as `/dev/sda5` and `PARTNAME` to `sda5`

At some point after this it goes off the rails a bit... I can manually mount the disk, move the files in `/var/lib/docker` over to `/mnt/sda5/` and rig it all up by hand.  I also noticed that I had to `umount /var/lib/docker` to remove the existing one properly, as anytime I did a `rm -rf /var/lib/docker` I got a busy error.

I realize a PR to fix the issue would be better here, but I'm out of time.  Figured someone might know how to address the issue.

Thanks!

PR #746, opened 2015-02-18
Improvements in the FAQ (persistence partitions and bootlocal.sh)

Vrakfall:
Adds some lines to the doc/FAQ.md

-Adds that a default persistence partition is made by the b2d ISO image
-Adds that 2 persistence partitions can't have the same name but
different labels

-Adds that /var/lib/boot2docker/bootlocal.sh script mus be in /bin/sh

### DIFF
--- a/doc/FAQ.md
+++ b/doc/FAQ.md
@@ -34,13 +34,21 @@ distributed systems. I just wanted the fastest way to boot to Docker.
 
 **Persistent partition choice**
 
-Boot2Docker will first try to mount a partition labeled ``boot2docker-data``, if
+Boot2Docker will first try to mount a partition named ``boot2docker-data``, if
 that doesn't exist, it will pick the first ``ext4`` partition listed by ``blkid``.
+
+By default, the ISO image creates the ``boot2docker-vm`` partition, therefore no additional one is likely to be needed but you can still add some if you want to.
+
+Keep in mind that if you have 2 partitions with the same names but different labels,
+boot2docker will only see one of them. So, if you add a ``boot2docker-data`` partition, boot2docker will mount that one and forget ``boot2docker-vm``.
+However, if you name them differently, boot2docker will see all of them.
+E.g.: ``boot2docker-vm`` and ``partition2`` will both be mounted.
 
 **Local Customisation (with persistent partition)**
 
 If you have a persistence partition, you can make customisations that are run at
 the end of the boot initialisation in the ``/var/lib/boot2docker/bootlocal.sh`` file.
+That script must be in ``#!/bin/sh`` because neither bash or csh are installed in the boot2docker vm.
 
 You can also set variables that will be used during the boot initialisation (after
 the automount) by setting them in `/var/lib/boot2docker/profile`
@@ -65,7 +73,7 @@ su - docker -c "tce-load -i /var/lib/boot2docker/*.tcz"
 docker run -d -v /var/run/docker.sock:/var/run/docker.sock $(which docker):$(which docker)  -name dom0 svens-dom0
 ```
 
-Or, if you need to tell the Docker daemon to use a specific DNS server, add the 
+Or, if you need to tell the Docker daemon to use a specific DNS server, add the
 following to ``/var/lib/boot2docker/profile``:
 
 ```


doc/FAQ.md
```diff
@@ -34,13 +34,21 @@ distributed systems. I just wanted the fastest way to boot to Docker.
...
-Boot2Docker will first try to mount a partition labeled ``boot2docker-data``, if
+Boot2Docker will first try to mount a partition named ``boot2docker-data``, if
 that doesn't exist, it will pick the first ``ext4`` partition listed by ``blkid``.
 
+By default, the ISO image creates the ``boot2docker-vm`` partition, therefore no additional one is likely to be needed but you can still add some if you want to.
+
+Keep in mind that if you have 2 partitions with the same names but different labels,
+boot2docker will only see one of them. So, if you add a ``boot2docker-data`` partition, boot2docker will mount that one and forget ``boot2docker-vm``.
+However, if you name them differently, boot2docker will see all of them.
+E.g.: ``boot2docker-vm`` and ``partition2`` will both be mounted.
```
tianon: I'm not at all sure what you mean by `name` versus `label` here. :confused:

doc/FAQ.md
```diff
@@ -34,13 +34,21 @@ distributed systems. I just wanted the fastest way to boot to Docker.
...
+
 **Local Customisation (with persistent partition)**
 
 If you have a persistence partition, you can make customisations that are run at
 the end of the boot initialisation in the ``/var/lib/boot2docker/bootlocal.sh`` file.
+That script must be in ``#!/bin/sh`` because neither bash or csh are installed in the boot2docker vm.
```
tianon: :+1:!

doc/FAQ.md
Vrakfall: Let's take `boot2docker-vm` as example. `boot2docker` would be the name part of it and `-vm` the label part. If we have 4 different data disk attached, respectively called `boot2docker-data`, `boot2docker-alpha`, `testname-one` and `testname-two`, boot2docker will attach `boot2docker-data`(because it is the first name it is looking for) then it will attach `testname-one` and will forget the two others because they have the same name.

Well, I didn't check this behaviour for days so I might be a little bit wrong. At least, I'm sure that `boot2docker-alpha` will be forgotten.

My above explanation should probably be in the doc, as it is a bit clearer I guess.

Feel free to change it as you wish to make it clearer, the English isn't my mother language.

doc/FAQ.md
Vrakfall: It took me at least 24 hours to figure it out. =D Thought it would save some time from others.

doc/FAQ.md
tianon: From looking at the script in question (https://github.com/boot2docker/boot2docker/blob/175488c34ff9c05c1307705c47a4d8a1282951b8/rootfs/rootfs/etc/rc.d/automount#L3-L8), I don't see how it would ever even attempt to mount any partition that doesn't have a label of `boot2docker-data` explicitly.  Now, if you've got two partitions with that same label, it's true that only the first one will be mounted, but I don't see that really as a problem we can reasonably solve since we're mounting based on partition label.

tianon: I suppose there's a fallback to just mount the first `ext4` or `btrfs` partition, but that's still not based on name or label.

Vrakfall: You're right. I can't find back the issue I was inspired from, they were telling something like I did. I did some tests and the only partition that seems to be mounted is `boot2docker-vm`.

Anyway, to fix my commit I need some more information. According to the script you linked, `boot2docker-vm` shouldn't even be loaded BUT according to my tests no partition can be loaded except `boot2docker-vm`. Even `boot2docker-data` isn't loaded if attached alone. Maybe there's a link with #339 but I didn't check much.

Vrakfall:
Hi, any update ?

matthughes:
I like what you added, but think we could have more docs here. 

1) I get that bootlocal.sh/bootsync.sh live in /var/lib/boot2docker/ but how do I *get* the files there?  Is the only way to ssh and manually edit them?  I'd love to put those files in ~/.boot2docker on my *host* and have them auto-sync when I init boot2docker.

2) Still confused about when each of these scripts run.  Does bootlocal.sh run before or after bootsync.sh?

PR #747, opened 2015-02-19
Proper VMware support

frapposelli:
As mentioned in #638 and #659 this PR adds support for VMware Tools in the most comprehensive manner given the current limitation of TCL 5.x.

Features:
- Add `vmtoolsd` for backdoor communication with the hosting hypervisor.
- The `open-vm-tools` build is embedded in the Dockerfile (no longer tied to [vmware/tcl-container](https://github.com/vmware/tcl-container)).
- A wrapper for the `shutdown` command, to gracefully shutdown and reboot the GOS when issuing commands using tools.
- Memory balloon kernel module (`vmw_balloon`).
- A slightly modified `automount` script that verifies if the disk is provided by a VMware hypervisor, throughly checks if empty and formats it for use with B2D.

Missing:
- Shared Folders (`vmhgfs`) support as we need 64bit userspace and kernel to make it work (will follow up with a PR on that too).

Fixes #638 and #659 

References:
docker/machine#507
docker/machine#532
docker/machine#461
docker/machine#430


SvenDowideit:
cool! how big does this make the iso?

and have you had a chance to test it with the 64 bit work from @afressancourt ?

romant:
+1

frapposelli:
@SvenDowideit size has not increased much:
```shell
$~ ls -al boot2docker.iso
-rw-r--r--  1 frapposelli  staff  25165824 Feb 19 11:31 boot2docker.iso
```
And yes, I have a 64 bit working install based on @afressancourt work that has everything including shared folders support, is there a 64 bit branch I can send that PR to?

afressancourt:
Hello Fabio,

Do you have the Dockerfile somewhere that I could build upon in parallel of the pullrequest process? Thanks a lot, and congratulations, I am struggling with shared folder support and acpid since the beginning of the week !

frapposelli:
@afressancourt I need to do some merging before pushing the branch to github but I'll update you once it's there.

BTW, when I say everything working including shared folders support it's in regard to the scope of this PR so VMware Tools, I haven't played with it on Virtualbox :smile:

afressancourt:
My own pull request 
https://github.com/boot2docker/boot2docker/pull/748

frapposelli:
@SvenDowideit @tianon have you guys had the chance to review this?

rootfs/rootfs/usr/local/bin/lsb_release
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+# just to make open-vm-tools happy, nad fake output
```
SvenDowideit: `nad fake` :)

SvenDowideit:
yeah, LGTM - though I've not had a system setup to test it. @tianon ?

tianon:
> ... (no longer tied to vmware/tcl-container).

We still are though, right? (since we're downloading resources from that repo)

I wasn't opposed to keeping that built externally especially since it keeps the diff here down to just the elements that matter for boot2docker explicitly, and I trust you guys to maintain it appropriately.

The approach here seems reasonably sane, although adding the full build here has turned it into a bit of a patch-bomb. :smile:

> Showing  8 changed files  with 754 additions and 2 deletions.

frapposelli:
@tianon on the ties to vmware/tcl-container, I will refactor it a bit and make it rely on a single patch file, I'd rather keep the build process as part of boot2docker, it is easier to maintain for us in the long run :smile:

tianon:
Heh OK, that's fair. :+1:

frapposelli:
Here it is, bundled the patches and did some general cleanup, most of the LOC come from the network script that we added :wink:

Dockerfile
```diff
@@ -150,6 +151,50 @@ RUN mkdir -p /vboxguest && \
...
     cp x86/lib/VBoxGuestAdditions/mount.vboxsf $ROOTFS/sbin/
 
+# Build VMware Tools
+ENV OVT_VERSION 9.4.6-1770165
+
+# Download and prepare ovt source
+RUN mkdir -p /vmtoolsd && cd /vmtoolsd && \
+    curl -L -o open-vm-tools-$OVT_VERSION.tar.gz http://downloads.sourceforge.net/open-vm-tools/open-vm-tools-$OVT_VERSION.tar.gz && \
+    tar xfz open-vm-tools-$OVT_VERSION.tar.gz && rm open-vm-tools-$OVT_VERSION.tar.gz &&\
+    mv open-vm-tools-$OVT_VERSION open-vm-tools
```
tianon: Why not use `--strip-components` on `tar` to avoid the need to move the directory afterwards?  Something like:
```Dockerfile
# Download and prepare ovt source
RUN mkdir -p /vmtoolsd/open-vm-tools \
    && curl -L http://downloads.sourceforge.net/open-vm-tools/open-vm-tools-$OVT_VERSION.tar.gz \
        | tar -xzC /vmtoolsd/open-vm-tools --strip-components 1
```

That way, we can even avoid the on-disk tar file completely.

Dockerfile
```diff
@@ -150,6 +151,50 @@ RUN mkdir -p /vboxguest && \
...
+
+# Apply patches
+RUN cd /vmtoolsd && \
+    curl -L -o open-vm-tools-3.x.x-patches.patch https://gist.github.com/frapposelli/5506651fa6f3d25d5760/raw/a2310970f1b010c57b25d69a5465aa03fe9788fe/open-vm-tools-3.x.x-patches.patch &&\
```
tianon: Pulled from a very explicit commit off your Gist -- love it. :heart_eyes:

tianon: Makes it a little hard to update later (by anyone who isn't you), but this is only temporary, right?

frapposelli: Yes, this is definitely temporary, I really wanted to put this gist under our github org but apparently it's not possible yet. In the future we're going to have a specific patches repo in our org.

Dockerfile
frapposelli: Makes sense, it's that I'm always a bit anxious when piping stuff out of curl :smile: let me fix that

Dockerfile
tianon: Oh, I thought the patch itself was going to be temporary -- maybe some description about why the patch is necessary in comments here would be warranted?

tianon: (so we don't repeat this conversation :+1:)

Dockerfile
```diff
@@ -150,6 +151,50 @@ RUN mkdir -p /vboxguest && \
...
+                                                                         libglib2.0-dev libdumbnet-dev:i386 \
+                                                                         libdumbnet1:i386 libfuse2:i386 libfuse-dev \
+                                                                         libglib2.0-0:i386 libtirpc-dev libtirpc1:i386
+
+RUN ln -s /lib/i386-linux-gnu/libglib-2.0.so.0.3200.4 /lib/i386-linux-gnu/libglib-2.0.so &&\
+    ln -s /lib/i386-linux-gnu/libtirpc.so.1.0.10 /lib/i386-linux-gnu/libtirpc.so &&\
+    ln -s /usr/lib/i386-linux-gnu/libgthread-2.0.so.0 /usr/lib/i386-linux-gnu/libgthread-2.0.so &&\
+    ln -s /usr/lib/i386-linux-gnu/libgmodule-2.0.so.0 /usr/lib/i386-linux-gnu/libgmodule-2.0.so &&\
+    ln -s /usr/lib/i386-linux-gnu/libgobject-2.0.so.0 /usr/lib/i386-linux-gnu/libgobject-2.0.so &&\
+    ln -s /lib/i386-linux-gnu/libfuse.so.2 /lib/i386-linux-gnu/libfuse.so
```
tianon: Sounds like somebody needs to fix the `open-vm-tools` build system! :smile:

Dockerfile
```diff
@@ -150,6 +151,50 @@ RUN mkdir -p /vboxguest && \
...
+    CC="gcc -m32" CXX="g++ -m32" ./configure --host=i486-pc-linux-gnu --build=i486-pc-linux-gnu \
+                --without-kernel-modules --without-pam --without-procps --without-x --without-icu &&\
+    make CC="gcc -m32" CXX="g++ -m32" LIBS="-ltirpc" CFLAGS="-Wno-implicit-function-declaration" &&\
+    make DESTDIR=$ROOTFS install
+
+ENV LIBDNET libdnet-1.11
+
+RUN cd /vmtoolsd &&\
+    curl -L -o ${LIBDNET}.tar.gz http://sourceforge.net/projects/libdnet/files/libdnet/${LIBDNET}/${LIBDNET}.tar.gz &&\
+    tar zxf ${LIBDNET}.tar.gz && rm ${LIBDNET}.tar.gz &&\
```
tianon: Same simple `tar` piping could be applied here :+1:

rootfs/rootfs/etc/rc.d/automount
```diff
@@ -34,6 +34,37 @@ if [ ! -n "$BOOT2DOCKER_DATA" ]; then
             mkfs.ext4 -L $LABEL $BOOT2DOCKER_DATA
             swapon "${UNPARTITIONED_HD}2"
         fi
+
+        DISK_VENDOR=`cat /sys/class/block/$(basename $UNPARTITIONED_HD /dev/)/device/vendor /sys/class/block/$(basename $UNPARTITIONED_HD /dev/)/device/model | tr -d "\n"`
```
tianon: Can we swap outer the `` ` ``s here for `$()`?

rootfs/rootfs/etc/rc.d/automount
```diff
@@ -34,6 +34,37 @@ if [ ! -n "$BOOT2DOCKER_DATA" ]; then
...
+        # Test if disk is "VMware, VMware Virtual S" and empty.
+        if [ "$DISK_VENDOR" = "VMware, VMware Virtual S" ]; then
+            # Check whether the disk has any known partitions on it
+            blkid -o device $UNPARTITIONED_HD
+            if [ $? == 2 ]; then
+                # As there are no partitions, let's make sure the disk is empty for real
+                dd if=$UNPARTITIONED_HD of=device_test_file bs=1k count=256 > /dev/null 2>&1
+                NON_NUL=$(<device_test_file tr -d '\0\n' | wc -c)
```
tianon: Nice -- love that you're going the extra mile to verify for sure that the disk is definitely zeroed out (at least for the first 256k).

Dockerfile
frapposelli: Original comment in my code was "horrible hack" :wink:

tianon: +++ would love to see that here too! :smile:

rootfs/rootfs/etc/rc.d/vmtoolsd
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+/usr/local/bin/vmtoolsd --background /var/run/vmtoolsd.pid
```
tianon: I feel like I've asked this before, but is there a nice easy way we can make this conditional based on whether we're really in VMware without hacky scraping of `/sys` or `/proc` or relying on the `vmtoolsd` binary to "be a good citizen"?  For example, both the VirtualBox-specific code and the HyperV-specific code have an `if modeprobe some-module ...` that works really, really well to make sure we don't even attempt to run the code unless we've got the necessary hardware we expect.

frapposelli: I'm positive that `vmtoolsd` would quit if can't talk to the vmx backdoor, but let me test it on VirtualBox to be 100% sure.

rootfs/rootfs/etc/vmware-tools/scripts/vmware/network
```diff
@@ -0,0 +1,596 @@
+#!/bin/sh
```
tianon: I hate that we've fragmented this discussion :cry: -- https://github.com/boot2docker/boot2docker/pull/638#issuecomment-66576677: is there any way we could just keep a patch to the `open-vm-tools` version instead of a full copy?  Better yet would be to figure out a way to make the original work, but I'll settle for this if we can have some clear way of saying "this is the original and this is what had to be changed for boot2docker", especially so that we can update as necessary later and not forget what was broken and had to be fixed here in the interim.

rootfs/rootfs/etc/rc.d/vmtoolsd
tianon: Thanks :heart:

If it is indeed a good citizen, let's at least add a comment noting as much
so that we know where the assumption was encoded if we manage to find a
situation it's not a good citizen in later. :+1:

rootfs/rootfs/etc/vmware-tools/scripts/vmware/network
frapposelli: Let me take a deeper look into this, I'm not even 100% sure this file is needed for boot2docker at all.

frapposelli: It is indeed needed by vmtoolsd, I will add it to the patch routine instead of adding the file directly to the rootfs.

rootfs/rootfs/etc/rc.d/vmtoolsd
frapposelli: It is a good citizen :smile: added a comment to note that, just tested with VirtualBox:

vmtoolsd in virtualbox

Dockerfile
frapposelli: Added a comment, it all boils down to the release cadence of `open-vm-tools`, the patch will go away once we have a new release that incorporates those changes.

frapposelli:
Alright, spit and polish all around, should be ready for prime time this time :smile:

tianon:
LGTM

How hard will it be to port to a proper 64bit userspace?

frapposelli:
@tianon has been done already :smile: I have a branch running at https://github.com/frapposelli/boot2docker/tree/vmware-64bit based on the work made by @afressancourt 

It's actually better as it supports shared folders (can't make them work with the current 64bit kernel/32bit userspace situation).

md5:
> It's actually better as it supports shared folders (can't make them work with the current 64bit kernel/32bit userspace situation).

:raised_hands:

rootfs/rootfs/etc/rc.d/automount
frapposelli: If there's nothing in the first 256k it's a very weird way to store data in a block device :wink:

frapposelli:
I just realized I left out an important bit from the auto mount script, need to fix it before merging.

frapposelli:
Fixed the automount script, forgot that we have a different SCSI VID and PID in ESX :-).

Tested with Fusion, Workstation, Player and ESX, so LGTM :-)

rootfs/rootfs/etc/rc.d/automount
```diff
@@ -34,6 +34,37 @@ if [ ! -n "$BOOT2DOCKER_DATA" ]; then
             mkfs.ext4 -L $LABEL $BOOT2DOCKER_DATA
             swapon "${UNPARTITIONED_HD}2"
         fi
+
+        DISK_VENDOR=$(cat /sys/class/block/$(basename $UNPARTITIONED_HD /dev/)/device/vendor /sys/class/block/$(basename $UNPARTITIONED_HD /dev/)/device/model | tr -d "\n")
+        # Test if disk is "VMware, VMware Virtual S" and empty.
+        if [ "$DISK_VENDOR" = "VMware, VMware Virtual S" ] || [ "$DISK_VENDOR" = "VMware  Virtual disk    " ]; then
```
tiborvass: @frapposelli I might be doing something weird (and I'm not testing your PR directly), but I wanted to check the vendor of my sda on VMWare Fusion 7, and it only said `VMware, `. Do you know why that would be the case?

tianon: Maybe this should just be `[[ "$DISK_VENDOR" == *VMware* ]]` instead?

dreampuf:
I'm exciting to see this PR get merge for shared folder feature.
Thank you for great job !

rootfs/rootfs/etc/rc.d/automount
tiborvass: +1 for `*VMware*`

rootfs/rootfs/etc/rc.d/automount
tianon: Indeed! :heart:

weikinhuang:
Has there been any updates on this?

dizzyd:
*bump* :)

mikeys:
Wondering about this myself... Currently it's very difficult to utilise boot2docker in development environment due to the slowness of VirtualBox's shared folders implementation.

rootfs/rootfs/etc/rc.d/automount
frapposelli: Apparently busybox doesn't support regex in tests, I can't make this to work...

```shell
docker@fusion-test-b2d:~$ cat regextest.sh
#!/bin/sh

DISK_VENDOR="asdasd VMware qweqweqwe"

if [[ "$DISK_VENDOR" == *VMware* ]]; then
        echo "yay"
else
        echo "nay"
fi

docker@fusion-test-b2d:~$ ./regextest.sh
nay
docker@fusion-test-b2d:~$
```

We should be fine with the dual test for now, I guess.

md5: I think it will work if you change the shebang to /bin/bash

frapposelli: @md5 bash is not part of boot2docker

md5: D'oh!

tianon: Ouch.  I didn't realize `[[`'s pattern-matching wasn't POSIX.  That hurts.

Here's a hack that _does_ work, though!
```console
$ docker run -it --rm busybox
/ # vmware='some VMware something'
/ # [ "${vmware%%*VMware*}" != "$vmware" ] && echo yes || echo no
yes
```

tianon:
Ok, this LGTM now; I'm planning to merge as soon as 1.6 is finished and released.

frapposelli:
Great, thanks @tianon !

tianon:
Hey thanks for the poke @frapposelli, and sorry we were already committed and frozen for 1.6 :disappointed:.  Onward to merge, though!

md5:
:tada:

cmluciano:
Are there any docs for using this within boot2docker? I tried changing my driver using boot2docker config to no avail.

PR #790, opened 2015-03-21
Nicer boot2docker ssh example

devinus:
This doesn't allocate a pseudo-tty and gives a nice foreground log of connections being made.

### DIFF
--- a/doc/WORKAROUNDS.md
+++ b/doc/WORKAROUNDS.md
@@ -11,7 +11,7 @@ your other computers on your LAN. You can do it temporarily, using `ssh`:
 Run following command (and keep it open):
 
 ```sh
-$ boot2docker ssh -L 8000:localhost:8000
+$ boot2docker ssh -vnNTL 8000:localhost:8000
 ```
 
 or you can set up a permanent VirtualBox NAT Port forwarding:


tianon:
LGTM

For reference:
```
$ man ssh
...
     -v      Verbose mode.  Causes ssh to print debugging messages about its
             progress.  This is helpful in debugging connection, authentica‐
             tion, and configuration problems.  Multiple -v options increase
             the verbosity.  The maximum is 3.
...
     -n      Redirects stdin from /dev/null (actually, prevents reading from
             stdin).  This must be used when ssh is run in the background.  A
             common trick is to use this to run X11 programs on a remote
             machine.  For example, ssh -n shadows.cs.hut.fi emacs & will
             start an emacs on shadows.cs.hut.fi, and the X11 connection will
             be automatically forwarded over an encrypted channel.  The ssh
             program will be put in the background.  (This does not work if
             ssh needs to ask for a password or passphrase; see also the -f
             option.)
...
     -N      Do not execute a remote command.  This is useful for just for‐
             warding ports (protocol version 2 only).
...
     -T      Disable pseudo-tty allocation.
...
```

PR #825, opened 2015-04-11
Enable CFS_BANDWIDTH to support cpu cfs quota

coolljt0725:
Signed-off-by: Lei Jitang <leijitang@huawei.com>

Enable CFS_BANDWIDTH to support cpu cfs quota


tianon:
LGTM

PR #835, opened 2015-04-17
Fixed a broken path for the docker.sock

far-blue:
I noticed after updating to the 1.6.0 docker iso that jwilder/nginx-proxy was failing. I've tracked this to a change in the unix socket path which I believe is due to a bug in the init script where the unix path is now set to unix://. This pull request just puts it back to unix:///var/run/docker.sock

### DIFF
--- a/rootfs/rootfs/usr/local/etc/init.d/docker
+++ b/rootfs/rootfs/usr/local/etc/init.d/docker
@@ -93,8 +93,8 @@ start() {
     ulimit -p $DOCKER_ULIMITS
 
     echo "------------------------" >> "$DOCKER_LOGFILE"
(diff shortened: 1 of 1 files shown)


md5:
What behavior are you seeing? Looking through the code, it looks like `unix://` should default to `/var/run/docker.sock` (cf. https://github.com/docker/docker/blob/7233bd223d4760af604e619a06c2c9ffe3fa1e22/api/common.go#L22)

far-blue:
On further investigation I think the unix socket issue is a red herring and is a symptom of an issue with restarting docker within docker-machine (i.e. restarting of the docker daemon, not restarting containers). I'll raise an issue against docker-machine with the details but to summarise it appears to be a possible race condition on the docker.sock which was causing docker to sometimes be able to create it - leading to random fails. My fix seemed to work but, of course, that's the nature of things like race conditions!

md5:
Thanks for the update @far-blue :+1: 

For those following along at home, the `machine` issue is here: https://github.com/docker/machine/issues/1027

PR #874, opened 2015-05-12
Fix install-ca-files to use basename, not path

rubbish:
I ran into this while trying out the instructions in the README. This change was required so that it will read the file correctly and handle the linking.

Fixes #873


rubbish:
#873 will fix this as well.

tianon:
Oh bollocks, thanks -- can you fix the indentation to match the surrounding block? (tabs vs spaces)

rubbish:
@tianon Updated.

tianon:
Thanks, LGTM :+1:

PR #889, opened 2015-05-26
Install kernel headers for Hyper-V KVP build to /tmp/kheaders

### LINKED ISSUE #826: Hyper-V KVP kernel headers_install in wrong location
The `make headers_install INSTALL_HDR_PATH=/usr` command in the Hyper-V KVP daemon part of the Dockerfile ([line 181](https://github.com/boot2docker/boot2docker/blob/master/Dockerfile#L181))  installs the kernel.org header files over the Debian header files already present in `/usr/include`.

This is obviously a bad idea because it creates a hodge-podge of debian and kernel.org headers.
PR to fix this is on the way.

tianon:
Closes #827
Fixes #826

@simonvanderveldt care to take a look and make sure I haven't misrepresented here?  (and to make sure this will actually work without the `sed` -- it builds successfully, but I'm not sure how to even verify #826, so I'm happy to replace this back to the `sed` if it turns out that doing this the same way we do for AUFS is not OK)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -117,7 +117,7 @@ RUN cd /linux-kernel && \
     git clone http://git.code.sf.net/p/aufs/aufs-util && \
     cd /aufs-util && \
     git checkout aufs4.0 && \
-    CPPFLAGS="-I/tmp/kheaders/include" CLFAGS=$CPPFLAGS LDFLAGS=$CPPFLAGS make && \
+    CFLAGS='-I/tmp/kheaders/include' CPPLFAGS="$CFLAGS" LDFLAGS="$CPPFLAGS" make && \
     DESTDIR=$ROOTFS make install && \
     rm -rf /tmp/kheaders
 
@@ -250,10 +250,11 @@ RUN cd $ROOTFS \
 
 # Build the Hyper-V KVP Daemon
 RUN cd /linux-kernel && \
-    make headers_install INSTALL_HDR_PATH=/usr && \
+    make INSTALL_HDR_PATH=/tmp/kheaders headers_install && \
     cd /linux-kernel/tools/hv && \
-    make hv_kvp_daemon && \
-    cp hv_kvp_daemon $ROOTFS/usr/sbin
+    CFLAGS='-I/tmp/kheaders/include' CPPLFAGS="$CFLAGS" LDFLAGS="$CPPFLAGS" make hv_kvp_daemon && \
+    cp hv_kvp_daemon $ROOTFS/usr/sbin && \
+    rm -rf /tmp/kheaders
 
 # These steps can only be run once, so can't be in make_iso.sh (which can be run in chained Dockerfiles)
 # see https://github.com/boot2docker/boot2docker/blob/master/doc/BUILD.md


tianon:
ping @simonvanderveldt (I'm not sure how to reproduce the issues your PR is solving, so I don't know how to test that this still solves them :cry:)

simonvanderveldt:
Sorry for the lack of response, bit busy here :)
As far as I can remember the Makefile from Hyper-V KVP didn't read the `CFLAGS` because it directly sets them within the Makefile. Could be wrong though and/or might have changed in the meantime.
Does the build work OK?

I can properly check it this weekend, is that soon enough?

simonvanderveldt:
@tianon it's indeed as I expected.
```
Step 45 : RUN cd /linux-kernel &&     make INSTALL_HDR_PATH=/tmp/kheaders headers_install &&     cd /linux-kernel/tools/hv &&     CFLAGS='-I/tmp/kheaders/include' CPPLFAGS="$CFLAGS" LDFLAGS="$CPPFLAGS" make hv_kvp_daemon &&     cp hv_kvp_daemon $ROOTFS/usr/sbin &&     rm -rf /tmp/kheaders
 ---> Running in 6d4510939ed1
  CHK     include/generated/uapi/linux/version.h
  INSTALL include/asm-generic (35 files)
   ...
  INSTALL include/asm (65 files)
gcc -Wall -Wextra -g -lpthread -o hv_kvp_daemon hv_kvp_daemon.c
```
This compared to my PR which uses `sed`
```
Step 42 : RUN cd /linux-kernel &&     make INSTALL_HDR_PATH=/tmp/kheaders headers_install &&     cd /linux-kernel/tools/hv &&     sed -i 's!\(^CFLAGS = .*\)!\1 -m32 -I/tmp/kheaders/include!' Makefile &&     make hv_kvp_daemon &&     cp hv_kvp_daemon $ROOTFS/usr/sbin &&     rm -rf /tmp/kheaders
 ---> Running in 92e7eaf357c9
  CHK     include/generated/uapi/linux/version.h
  INSTALL include/asm-generic (35 files)
  ...
  INSTALL include/asm (65 files)
gcc -Wall -Wextra -g -lpthread -m32 -I/tmp/kheaders/include -o hv_kvp_daemon hv_kvp_daemon.c
```
The `-I/tmp/kheaders/include` is missing from the `gcc` command. Which means gcc isn't using the correct kernel headers.
(Since we've moved to 64-bit the `-m32` obviously isn't necessary anymore)

This happens because the [KVP Makefile](https://git.kernel.org/cgit/linux/kernel/git/stable/linux-stable.git/tree/tools/hv/Makefile?id=refs/tags/v4.0.4#n6) assigns/overwrites `CFLAGS`:
```
CFLAGS = $(WARNINGS) -g $(PTHREAD_LIBS)
```
If they would've used `+=` we would've been able to set `CFLAGS` from the command line. I guess that's something that should be fixed upstream.

This was the reason I used `sed` to replace the CFLAGS value in the Makefile directly.

If you want I can rebase my PR on current master.

tianon:
Sure, if you'd like to rebase yours that'd be excellent!  I'll just close this, since it's clearly a failed attempt. :+1:

PR #893, opened 2015-05-28
Add CONFIG_IOSCHED_CFQ

tianon:
From `check-config.sh` (for optional features). :+1:


tianon:
(tested and build is working)

PR #898, opened 2015-06-01
Allow large core files by default

matschaffer:
Just started doing this on my docker hosts in the cloud.

Seemed like something worth having on boot2docker-based dev environments as well.


tianon:
While I agree this can be useful, I think it'd be more appropriate to either propose this as the default upstream in Docker itself, or to document how users can add `EXTRA_ARGS` to `/var/lib/boot2docker/profile` in order to achieve this.  We try to keep changes to "the Docker experience" as minimal as possible.

PR #924, opened 2015-06-10
Enable CONFIG_FTRACE

etown:
I work on a tool ( https://github.com/draios/sysdig ) that provides excellent visibility into containers and it would be nice if it could work with boot2docker. Unfortunately tracepoints aren't fully enabled in the kernel configuration, which is surprising since it's harmless and have been enabled by default on almost every linux distribution released in the last 6 years. Thank you.


tianon:
Seems sane to me. :+1:

LGTM

simonvanderveldt:
FYI This has been enabled on Debian since at least Squeeze/6.0 from 2011 :)

PR #927, opened 2015-06-11
Fix show not unicode file(like chinese) when -v mount folder.

wufeifei:
Change mount to mount.vboxsf


simonvanderveldt:
This is a duplicate of #697

wufeifei:
OK

simonvanderveldt:
@wufeifei it might be useful to join the conversion @ #697, the change you proposed in this PR from `mount -t vboxsf` to `mount.vboxsf` is discussed there.

PR #931, opened 2015-06-12
Docs: Make VM data persistence (or lack of) clearer

### LINKED ISSUE #912: Docs: Make it clearer that only /var/lib/{boot2docker,docker} is persisted after reboot
On Windows, using boot2docker 1.6.2, when I perform a ``boot2docker.exe restart``, the state of the VM isn't persisted (eg my modifications to .ashrc are gone, as is the python/pip/docker-compose install that I added).

This behaviour seems unexpected (it should just be doing a reboot, right?), and isn't mentioned on the boot2docker.exe help page:
```
   restart             Gracefully reboot the VM.
```

Is this intentional, or a bug?

edmorley:
Previously it was easy to not realise that data outside of `/var/lib/docker` and `/var/lib/boot2docker` would be lost after reboots of the VM. We now also link to the relevant FAQ section from the main README, so it's easier to work out how to make permanent modifications.

Fixes #912.

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,10 +273,16 @@ pass: tcuser
 
 #### Persist data
 
+Boot2docker uses [Tiny Core Linux](http://tinycorelinux.net), which runs from
+RAM and so does not persist filesystem changes by default.
+
 When you run `boot2docker init`, the `boot2docker` tool auto-creates a disk that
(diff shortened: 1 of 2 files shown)


tianon:
LGTM

Issue #603, opened 2014-10-28
Pre-install docker-compose (useful on Windows)


josephpage:
Pre-install Fig is the only way I found to get a working Fig out of box with Boot2docker on Windows.
Fig is lighweight, very useful and now part of Docker, so... why not ?

I'm waiting for your comments.

mikehaertl:
Now that `docker-compose` is released, this issue should be updated to include that instead.

dduportal:
Just to let you know that this PR : https://github.com/boot2docker/boot2docker/pull/748 maybe a big help for the future : it will make boot2docker a full 64 Bits OS (user and kernel), leaving possibility to embed docker-compose

ebuildy:
+1, docker-compose is really missing

VonC:
@ebuildy but it does run just fine within a container, so it might not *really* be missing on Windows: see [How to install docker-compose on Windows](http://stackoverflow.com/a/29728993/6309).

edmorley:
Whilst the native Compose binary doesn't work in the boot2docker VM due to docker/compose#599, installing the [docker-compose Python package](https://pypi.python.org/pypi/docker-compose) inside boot2docker works perfectly (from what I can tell so far).

In fact, it seems to work even better than the "[docker-compose in a container](https://github.com/dduportal-dockerfiles/docker-compose)" method - since that currently has a caching bug (dduportal-dockerfiles/docker-compose#6).

As such for now, running the Python docker-compose package inside boot2docker seems to be the most reliable solution for Windows users (having spent many hours trying to battle with the alternatives).

To install docker-compose from PyPI, run this from inside boot2docker:
```bash
docker@boot2docker:~$ tce-load -wi python && curl https://bootstrap.pypa.io/get-pip.py | sudo python - && sudo pip install -U docker-compose
```

To save having to run the above every time the boot2docker VM is restarted (since changes don't persist), you can use bootlocal.sh like so:
```bash
docker@boot2docker:~$ echo 'su docker -c "tce-load -wi python && curl https://bootstrap.pypa.io/get-pip.py | sudo python - && sudo pip install -U docker-compose"' | sudo tee /var/lib/boot2docker/bootlocal.sh > /dev/null && sudo chmod +x /var/lib/boot2docker/bootlocal.sh
```
[...]

PR #932, opened 2015-06-14
Embed docker-compose inside boot2docker

dduportal:
Hello dear maintainers !

Since we have a lot of people dealing with boot2docker + virtual box portability issues (Windows essentially, but also vbox sharing folders and filesystem homogeneity between docker and compose client with the remote system), here is a little proposal to provide compose as a full part inside boot2docker.

Cons :
* Add a 5mb overhead :
```bash
docker@boot2docker:~$ docker-compose --version
docker-compose 1.2.0
docker@boot2docker:~$ du -sh /usr/local/bin/docker-compose 
5.0M	/usr/local/bin/docker-compose
```
* Docker-compose version is pretty tied to boot2docker releases (seems acceptable at first sight since you have a pretty good release monthly cycle)

Pros :
* Does not tie the end users to the use of compose in remote mode
* Let people use compose while you hard work on volumes, compose, boot2docker and all other hot stuff for this summer

Advices ?

For reference, here is a set of related issues/discussions on that topic (sendfile + vbox, windows support, docker volume path) :
* https://github.com/boot2docker/boot2docker/issues/603
* https://github.com/docker/compose/issues/599
* https://github.com/docker/compose/issues/1085

Thanks for your time !

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -225,7 +225,7 @@ RUN curl -L -o $ROOTFS/usr/local/bin/docker https://get.docker.io/builds/Linux/x
     { $ROOTFS/usr/local/bin/docker version || true; }
 
 # Get the docker-compose binary
-ENV COMPOSE_VERSION 1.2.0
+ENV COMPOSE_VERSION 1.3.1
 
 RUN curl -L -o $ROOTFS/usr/local/bin/docker-compose https://github.com/docker/compose/releases/download/${COMPOSE_VERSION}/docker-compose-Linux-x86_64 && \
     chmod +x $ROOTFS/usr/local/bin/docker-compose && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -224,6 +224,13 @@ RUN curl -L -o $ROOTFS/usr/local/bin/docker https://get.docker.io/builds/Linux/x
     chmod +x $ROOTFS/usr/local/bin/docker && \
     { $ROOTFS/usr/local/bin/docker version || true; }
 
+# Get the docker-compose binary
+ENV COMPOSE_VERSION 1.2.0
+
+RUN curl -L -o $ROOTFS/usr/local/bin/docker-compose https://github.com/docker/compose/releases/download/${COMPOSE_VERSION}/docker-compose-Linux-x86_64 && \
+    chmod +x $ROOTFS/usr/local/bin/docker-compose && \
+    $ROOTFS/usr/local/bin/docker-compose --version
+
 # Get the git versioning info
 COPY .git /git/.git
 RUN cd /git && \


simonvanderveldt:
@dduportal isn't it enough to be able to run compose inside a container?
Afaik with the recent fixes your image works as it should, right?

dduportal:
Hello !
You're right when talking about "advanced" user whom masters the data volumes.

In that case, target is "beginners" and "basics" users if docker who has not yet (or do not need) to think about complex filesystems path and bindings :)

Also, since there is no "official" docker-compose image, i'm not fan of pushing a personnal one : it does not bring enough quality insurance on my mind.

This was the philosophy of this PR, but maybe it will add more confusion also.

But anyway it is just a proposal. Thank for your time !

josephpage:
Great !!
:+1:  :+1: :+1: :+1: :+1: :+1: :+1:

tinchou:
As a total beginner that just wants to set up a dev environment in its Windows machine, I'd love to see docker-compose included by default

dduportal:
Updated to take in account compose 1.3.1 (https://github.com/docker/compose/releases/tag/1.3.1)

tianon:
@dduportal I like the idea, but I don't think I'm comfortable including this as-is in the base OS, and would prefer that we document the easiest way for users to do this (especially via `bootlocal.sh` for persistence)

Shouldn't users usually be running `docker-compose` directly on their host instead of inside the VM?

lmakarov:
@tianon 
>Shouldn't users usually be running docker-compose directly on their host instead of inside the VM?
This will be true and make this PR irrelevant as soon as `docker-compose` gets ported to Windows... :)

dduportal:
@tianon Yep, you're right, as @lmakarov stated, docker-compose should be run from the client outside boot2docker :)

This is why I propose this PR : for now, since docker-compose for Windows is not a reality, it will help user to begin with, without suffering :)

Another thing is the volume sharing which is a pain. The new Docker plugins may help, but the conceptual problem the user have still exists : I want to share the volumes from C:\bla to /app in the data container".

The idea of the PR will help user to begin softly, with a docker-engine path visibility, again the time that -v will be fully supported.

When the 2 things would have been implemented, i'll be the first the PR to remove the docker-compose binary :
* For achieving full transparent hypervisor concept with a ligthweight boot2docker
* Push user to understand the volume implications
* Follow you, Docker maintainers, state of mind by using client / side Docker that take advantages of API

Anyway, thank for your time :)

josephpage:
Embed docker-compose inside boot2docker may be a way to solve the inequality between OS X/Linux and Windows in the brand new (and awesome) [Docker Toolbox](https://github.com/docker/toolbox/blob/master/README.md).

dduportal:
Just a note here : since docker-compose is now supported in Windows, it does not make sense to pre-install inside boot2docker now :)

So : closing the MR !

josephpage:
Yes ! :+1:

ghahramani:
@dduportal Hi, now docker-compose supports all operating system, so I think it's good time to add it to boot2docker

you can check it here
https://docs.docker.com/compose/install/

dduportal:
Hello @ghahramani , given:

- The future of boot2docker (that should be dumped in the future, or change a lot)
- The stabilization of docker-compose patterns uses (e.g. using docker-compose and docker-cli outside boot2docker)

I closed the PR because it did not make any sense, especially regarding the time investment to maintain compose version while it is already packaged on docker-toolbox.

In your case, if you strongly need docker-compose inside boot2docker, you should build you own boot2docker to ensure control of this behavior.

PR #937, opened 2015-06-18
Update Dockerfile to be explicit about dependencies

tianon:
Let's get a tiny bit more consistency and repeatability here. :+1:


tianon:
Builds successfully! :balloon:

tianon:
Hmm, this could use a minor reordering to improve cache, too....

PR #963, opened 2015-06-29
Update kernel to 4.0.6 (and AUFS to v4.0-20150622)

tianon:


tianon:
Everything compiles and Docker works properly. :+1:

tianon:
Aaaaand https://lwn.net/Articles/649742/

PR #970, opened 2015-07-04
Fix setting swap partition type during automount

simonvanderveldt:
The sequence `(echo t; echo 2; echo 82; echo w) | fdisk $UNPARTITIONED_HD` to set the partition type for the swap partition to 82/Linux swap in the `/etc/rc.d/automount` script currently fails to do so because it doesn't match the commands that actually have to be entered.

Apparently when there's only 1 partition fdisk already selects this partition:
```
# I ran this manually after the initial
# (echo n; echo p; echo 2; echo ; echo +1000M ; echo w) | fdisk $UNPARTITIONED_HD

fdisk /dev/sda
...
Command (m for help): t
Selected partition 2
Hex code (type L to list codes):
```

No clue how long this hasn't been working correctly, but I couldn't find anything in the busybox release notes mentioning a change in fdisk behaviour related to this.

`fdisk -l` pre this fix
```
boot2docker ssh fdisk -l

Disk /dev/sda: 104.8 GB, 104857600000 bytes
255 heads, 63 sectors/track, 12748 cylinders
Units = cylinders of 16065 * 512 = 8225280 bytes

   Device Boot      Start         End      Blocks  Id System
/dev/sda1             124       12748   101410312+ 83 Linux
/dev/sda2               1         123      987966   2 Unknown

Partition table entries are not in disk order
```

`fdisk -l` after this fix
```
boot2docker ssh fdisk -l

Disk /dev/sda: 104.8 GB, 104857600000 bytes
255 heads, 63 sectors/track, 12748 cylinders
Units = cylinders of 16065 * 512 = 8225280 bytes

   Device Boot      Start         End      Blocks  Id System
/dev/sda1             124       12748   101410312+ 83 Linux
/dev/sda2               1         123      987966  82 Linux swap

Partition table entries are not in disk order
```

### DIFF
--- a/rootfs/rootfs/etc/rc.d/automount
+++ b/rootfs/rootfs/etc/rc.d/automount
@@ -28,7 +28,7 @@ if [ ! -n "$BOOT2DOCKER_DATA" ]; then
             (echo n; echo p; echo 2; echo ; echo +1000M ; echo w) | fdisk $UNPARTITIONED_HD
             # Let kernel re-read partition table
             partprobe
-            (echo t; echo 2; echo 82; echo w) | fdisk $UNPARTITIONED_HD
+            (echo t; echo 82; echo w) | fdisk $UNPARTITIONED_HD
             # Let kernel re-read partition table
             partprobe
             mkswap "${UNPARTITIONED_HD}2"


tianon:
Ok, now I'm worried.  This was "fixed" in #761, and this is essentially a revert of that.

simonvanderveldt:
@tianon Yeah, strange thing. I noticed that commit as well, so I triple checked everything but it really is not working correctly right now and I couldn't find a reason why it would be different now compared to when #761 was merged.

The only thing I can think of is that somehow something within TinyCore could have changed this behaviour. Maybe they switched from `fdisk` from util-linux to the one from busybox.
It could also have to do with the switch to TinyCore 6.0/64-bit for boot2docker.
I couldn't find a decent history/changelog for TinyCore though, so no way to check.

I'm building the iso from commit 6f2713b that merged #761 right now to check if it worked back then.

simonvanderveldt:
@tianon I just rebuilt the boot2docker iso for both commit 6f2713b (the commit after the merge of #761) and commit 175488c (the commit before @AntonioMeireles added his fix) and tested it with both of them + with the iso from boot2docker 1.5.0.

All of them them automatically select partition 2 when you want to change the partition type as I mentioned in the description of this PR.

Since #761 also added the `w` to write the changes to the swap partition to disk the actual effect of #761 is that the partition type for the swap partition was changed from the default `83/Linux` to `2/Unknown` because of the error in the `echo` line.

In other words the changes in #761 didn't actually fully fix the creation of the correct partition type.

tianon:
@simonvanderveldt awesome, thanks for doing the digging on this! :+1: :heart:

LGTM

PR #1034, opened 2015-08-12
use new 'daemon' command

gesellix:
Replace the deprecated `-d` command line flag with the new command name `daemon`.
See https://github.com/docker/docker/pull/13771 for details.

### DIFF
--- a/rootfs/rootfs/usr/local/etc/init.d/docker
+++ b/rootfs/rootfs/usr/local/etc/init.d/docker
@@ -99,8 +99,8 @@ start() {
     ulimit -p $DOCKER_ULIMITS
 
     echo "------------------------" >> "$DOCKER_LOGFILE"
-    echo "/usr/local/bin/docker -d -D -g \"$DOCKER_DIR\" -H unix:// $DOCKER_HOST $EXTRA_ARGS >> \"$DOCKER_LOGFILE\"" >> "$DOCKER_LOGFILE"
-    /usr/local/bin/docker -d -D -g "$DOCKER_DIR" -H unix:// $DOCKER_HOST $EXTRA_ARGS >> "$DOCKER_LOGFILE" 2>&1 &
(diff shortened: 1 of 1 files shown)


josephpage:
:+1:

tianon:
Awesome, thanks! :metal:

gesellix:
:bow:

md5:
This appears to have broken Unix socket connections to the Docker daemon.

I still see `-H unix://` in the command line used to call `docker daemon`, but it looks like that no longer causes the daemon to listen on `/var/run/docker.sock`:

```console
$ pgrep -d, docker | xargs ps -fp | cat
UID        PID  PPID  C STIME TTY          TIME CMD
root       892     1  0 03:35 ?        00:00:07 /usr/local/bin/docker daemon -D -g /var/lib/docker -H unix:// -H tcp://0.0.0.0:2376 --label provider=virtualbox --tlsverify --tlscacert=/var/lib/boot2docker/ca.pem --tlscert=/var/lib/boot2docker/server.pem --tlskey=/var/lib/boot2docker/server-key.pem -s aufs
```

This has broken `jwilder/nginx-proxy` and other containers that expect to connect via the Unix socket in their default configuration.

```console
docker@dev:~$ ls -l /var/run/docker.sock
srw-rw----    1 root     docker           0 Sep 12 03:35 /var/run/docker.sock
docker@dev:~$ fuser /var/run/docker.sock
docker@dev:~$ 
```

md5:
I opened a PR in https://github.com/boot2docker/boot2docker/pull/1059

This probably also needs a bug opened against `docker/docker` for the regression of no longer supporting `unix://` as a value for `-H`...

md5:
Nevermind... For whatever reason, `docker version` was saying this:

> Cannot connect to the Docker daemon. Is 'docker -d' running on this host?

But it doesn't appear to be an actual regression in the handing of `-H unix://`. Not sure what the actual cause was, but I saw a `docker daemon` pid running after a `docker-machine upgrade` and nothing was able to connect to the Unix socket. TLS connections from my host were working though.

PR #1035, opened 2015-08-17
ntpd now updates system time regardless of offset

### LINKED ISSUE #840: Boot2docker's VM time lagged behind the host's
Related to https://github.com/boot2docker/boot2docker/issues/290 (trigger b2d to run ntpdate on OSX/WIndows resume events)

I wonder if the above issue 290 has been fixed. I found that after sleeping and resuming my Mac, the time in boot2docker VM was 15min behind the host. I'm on the latest version (1.5.0).

mikesherov:
Fixes #840

From the `ntpd` man page: 
-g: Normally, ntpd exits with a message to the system log if the offset exceeds the panic threshold, which is 1000 s by default. This option allows the time to be set to any value without restriction; however, this can happen only once. If the threshold is exceeded after that, ntpd will exit with a message to the system log. This option can be used with the -q and -x options. See the tinker command for other options.

### DIFF
--- a/rootfs/rootfs/etc/rc.d/ntpd
+++ b/rootfs/rootfs/etc/rc.d/ntpd
@@ -12,7 +12,7 @@ if [ -n "$NTP_SERVER" ]; then
 		fi
 	done
 	
-	ntpd -d -n -p $NTP_SERVER > /var/lib/boot2docker/log/ntpd.log 2>&1 &
+	ntpd -d -g -n -p $NTP_SERVER > /var/lib/boot2docker/log/ntpd.log 2>&1 &
 else
 	echo 'NTP_SERVER not set; skipping starting ntpd' > /var/lib/boot2docker/log/ntpd.log
 fi


tianon:
LGTM :+1:

tianon:
Oh, this is actually a problem for our `ntpd`:
```console
docker@boot2docker:~$ sudo ntpd --help
BusyBox v1.23.1 (2015-02-22 15:53:49 UTC) multi-call binary.

Usage: ntpd [-dnqNw] [-S PROG] [-p PEER]...

NTP client/server

	-d	Verbose
	-n	Do not daemonize
	-q	Quit after clock is set
	-N	Run at high priority
	-w	Do not set time (only query peers), implies -n
	-S PROG	Run PROG after stepping time, stratum change, and every 11 mins
	-p PEER	Obtain time from PEER (may be repeated)
		If -p is not given, read /etc/ntp.conf

```

mikesherov:
Sorry for that @tianon

tianon:
No worries, glad we caught it! :smile:

mikesherov:
@tianon is `-q` what I was looking for instead?

tianon:
no, I think `-q` is the equivalent of the `ntpdate` command; set the clock just once and then stop the service

PR #1044, opened 2015-08-26
Use iocharset=utf8 for automunt-shares

niw:
According to the document, it said that iocharset is default to utf8, however, actual behavior is not.
Force using utf8 to address the issue that non-ascii characters files disapper from the shared folders.

This is basically same as boot2docker/boot2docker#697 which is not updated since May, but doesn't change `mount -t vbox` to `mount.vbox` also keeps `defaults` option.

This will solve the issue similar to docker/machine#1763.

Closes boot2docker/boot2docker#697


zchee:
+1
Seems to troubled the Japanese user and Chinese user.

niw:
Just bump up for the patch CC: @tianon

tianon:
LGTM, thanks :+1:

PR #1047, opened 2015-09-02
automated_script.sh is written by ash for enable kernel option script

YujiOshima:
Rewrite automated_script.sh by ash.
Karnel option "script" is useful to install on bare metal server.
`if [[ $(tty) == "/dev/tty1" ]]; then`
looks unnecessary for boot2docker.
As in arch linux .automated_script.sh is called by .zlogin, that guard is necessary for arch linux.
https://projects.archlinux.org/archiso.git/commit/?id=d552ad32ab7001766ec2cdf6d2c555e91267a7cb

### DIFF
--- a/rootfs/rootfs/bootscript.sh
+++ b/rootfs/rootfs/bootscript.sh
@@ -83,8 +83,7 @@ if [ -e /var/lib/boot2docker/bootlocal.sh ]; then
 fi
 
 # Execute automated_script
-# disabled - this script was written assuming bash, which we no longer have.
-#/etc/rc.d/automated_script.sh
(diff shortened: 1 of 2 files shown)


tianon:
Sorry for the long tail!

Going to repeat my comment from https://github.com/boot2docker/boot2docker/issues/1276#issuecomment-333663786, namely:

> we should change the docs -- boot2docker is in pretty much just maintenance mode.  We're happy to accept bugfixes and simple PRs (especially related to the primary usage with `docker-machine`), but any additional functionality outside that is probably out-of-scope (and we'd especially like to keep the already-fragile boot process as simple as possible).

Thanks for the contribution!

Issue #1045, opened 2015-08-26
Build a boot2docker matching the docker experimental branch


Crazee:
docker-machine (et.al. using boot2docker) created machines are not happy if you're trying the experimental version of docker -- complains about API mismatch:

$ docker version
Client:
 Version:      1.9.0-dev
 API version:  1.21
 Go version:   go1.4.2
 Git commit:   1425b18
 Built:        Fri Aug 21 18:51:46 UTC 2015
 OS/Arch:      darwin/amd64
 Experimental: true
Error response from daemon: client is newer than server (client API version: 1.21, server API version: 1.20)

It would be very helpful if all the parts fit together.

md5:
@Crazee See #1057

PR #1057, opened 2015-09-10
Update Dockerfile.test to build from experimental

md5:
This PR updates the old `Dockerfile.test` file to build from `experimental.docker.com` instead of `test.docker.io`. This may address the request in #1045, though I think a `boot2docker/boot2docker:experimental` tag on Docker Hub would be needed to completely address that issue. Unfortunately, Docker Hub doesn't yet support named `Dockerfile`s, nor does it support building dependent Docker images reliably from a single `git push`.


md5:
Also, I think that `curl -f` should be used for most of the `curl` commands in these `Dockerfile`s, but I didn't make that change in this PR.

tianon:
LGTM

tianon:
(also, +1 to `-f`!)

PR #1059, opened 2015-09-12
Ensure daemon listens on /var/run/docker.sock

md5:
This fixes a breaking change from #1034 that prevents running `docker` commands locally in the boot2docker shell or using any container that mounts the default Docker Unix domain socket for its functionality (e.g. `jwilder/nginx-proxy`).


md5:
Actually, this might not be a regression... I just stopped and started the daemon again without this change and it appears to be working now...

PR #1064, opened 2015-09-16
Add initial .travis.yml

tianon:
Hopefully this builds fast enough (and has short enough output) for this to actually work well. :+1:


tianon:
Oh man, it's green.  Let's merge it before something happens!

tianon:
`ran for 35 min 26 sec` dat CI

PR #1067, opened 2015-09-25
Add OpenNebula support to boot2docker

jmelis:
It implements the following:
- Configure the network based on the OpenNebula Context.
  To do so, it calls the same script called by udhcpc passing
  the network configuration attributes.
- Small fix to automount script in order to properly handle
  disks generated by OpenNebula.
- Copy the SSH Key to the authorized_keys file of the docker
  user.

### DIFF
--- a/rootfs/rootfs/bootscript.sh
+++ b/rootfs/rootfs/bootscript.sh
@@ -99,3 +99,6 @@ fi
 
 # Load Parallels Tools daemon
 /etc/rc.d/prltoolsd
+
+# Launch OpenNebula
+/etc/rc.d/opennebula-context
--- a/rootfs/rootfs/etc/rc.d/automount
+++ b/rootfs/rootfs/etc/rc.d/automount
@@ -41,9 +41,10 @@ if [ ! -n "$BOOT2DOCKER_DATA" ]; then
             swapon "${UNPARTITIONED_HD}2"
         fi
 
+        IS_OPENNEBULA=$(blkid | grep -F 'LABEL="CONTEXT"')
         DISK_VENDOR=$(cat /sys/class/block/$(basename $UNPARTITIONED_HD /dev/)/device/vendor /sys/class/block/$(basename $UNPARTITIONED_HD /dev/)/device/model | tr -d "\n")
-        # Test if disk is "VMware, VMware Virtual S" and empty.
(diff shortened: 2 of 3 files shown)


jmelis:
Can anyone confirm that this is an error of the PR per se? or is there another cause? It looks like this is the problem:

    [91mE[0mFetched 118 MB in 6s (18.8 MB/s)
    [91m: [0m[91mFailed to fetch http://httpredir.debian.org/debian/pool/main/liba/libarchive-extract-perl/libarchive-extract-perl_0.72-1_all.deb  Error reading from server. Remote end closed connection [IP: 128.31.0.66 80]

    E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
    [0mThe command '/bin/sh -c apt-get update && apt-get -y install  unzip                         xz-utils                         curl                         bc                         git                         build-essential                         golang                         cpio                         gcc libc6 libc6-dev                         kmod                         squashfs-tools                         genisoimage                         xorriso                         syslinux                         isolinux                         automake                         pkg-config                         p7zip-full' returned a non-zero code: 100
    travis_time:end:10585dc7:start=1443200046099539261,finish=1443200072841985316,duration=26742446055
    [0K
    [31;1mThe command "docker build -t boot2docker/boot2docker ." exited with 1.[0m

Maybe it's just a spurious error...

jmelis:
Hi, any comments?

tianon:
I have some pretty hefty mixed feelings about this one.  Looking at OpenNebula, it appears to be for more production-grade cloud deployments.  The current "boot2docker" is definitely not a suitable operating system for a true production deployment as it stands today (and hence I'd hesitate to officially support much more than "local" deployment methods), however the future Debian-based boot2docker reimagining will be a better candidate for such a thing IMO (https://github.com/docker/boot2docker).

jmelis:
We are planning on adding support to the new boot2docker appliance as well, but for the time being I don't think it would be problematic having support for OpenNebula in this version of boot2docker. We could add a disclaimer noting your above comment for OpenNebula users, so this appliance is not used in production.

tianon:
Sorry for the long tail!

I'm still pretty strongly -1 on adding this extra support/complexity, especially since these days there are quite a few alternative distributions which are much better suited to the use case, such as RancherOS or even something purpose-built via LinuxKit (which is what Docker for Mac and Docker for Windows use to build their minimal Linux VM distro).

Thanks for the contribution! :heart:

PR #1073, opened 2015-10-12
Add /etc/timezone and /etc/localtime specifically for bind-mounting oneliners

### LINKED ISSUE #891: fix timezone and localtime
Currently the directories `/etc/localtime/  and /etc/timezone/` exist on the system.
It would be perfect if they would be files and not directories even if they are empty.

What I currently do is this
```
#!/bin/sh
sudo rm -rf /etc/localtime;
sudo rm -rf /etc/timezone;
echo "Europe/Berlin" | sudo tee /etc/timezone > /dev/null;
sudo touch /etc/localtime;
``` 

This is somehow related to #476 and https://github.com/docker/docker/issues/12811

tianon:
Fixes #891


Moghedrin:
LGTM

PR #1074, opened 2015-10-13
Hotfix: Parallels kernel modules are unavailable in v1.8.3

legal90:
Due to #1054, the `extra` subdirectory is no longer created. I was too naive while putting Parallels kernel modules to this dir :-(

So, in v1.8.3 there are no kernel modules from Parallels Tools, which causes a critical issue with shared folder mounting: https://github.com/Parallels/docker-machine/issues/7

In this PR I've changed the path for storing kernel modules from Parallels: let them be in the top level dir, adjacent to vbox modules.

@tianon Please, review this PR. We looking forward to get it released ASAP :pray:


tianon:
Arg :cry:

LGTM

legal90:
@tianon Thanks! Is there some estimation for the next release date?

tianon:
https://github.com/tianon/boot2docker-legacy/releases is where the RCs of 1.9.0 are dropping, which is going to be released next week :+1:

PR #1075, opened 2015-10-14
Update Parallels Tools to 11.0.2-31348

legal90:
There are no critical changes, but now this version could be built on Linux 4.2


tianon:
Niiiice, LGTM :+1:

All that's left now is AUFS support, eh? :smile:

legal90:
Yep )

PR #1081, opened 2015-10-21
fsck partition before mount

qxo:
we found filesystem  error on some boot2docker vms (on esxi)

when boot2docker is up, cannot umount the boot2docker-data device

if would be great for check and fix the issue on boot:)


tianon:
Hmm, this seems like a good idea, but my concern is that for larger disks it could take a really, really long time, and your VM is just inexplicably inaccessible until it finishes. :disappointed:

qxo:
Or we can do some tricky code, eg: cp /var/lib/boot2docker/bootsync.sh  to /tmp before run, then we can mount and fsck disk ,remount ...

PR #1088, opened 2015-11-02
Mount all VirtualBox shared folders by listing them with VBoxControl

u-foka:
It seems like VBoxControl does work with the current version, so all I had to do is listing the shared folders with it, and mount them.

Related to: https://github.com/docker/machine/pull/3798

### DIFF
--- a/rootfs/rootfs/etc/rc.d/vbox
+++ b/rootfs/rootfs/etc/rc.d/vbox
@@ -33,15 +33,8 @@ if modprobe vboxguest &> /dev/null && modprobe vboxsf &> /dev/null; then
 		return 0
 	}
 	
-	# bfirsh gets all the credit for this hacky workaround :)
-	try_mount_share /Users 'Users' \
-		|| try_mount_share /Users \
-		|| try_mount_share /c/Users 'c/Users' \
-		|| try_mount_share /c/Users \
-		|| try_mount_share /c/Users 'c:/Users' \
-		|| true
-	# TODO replace this whole hacky bit with VBoxService --only-automount
-	# (the problem with that being that we can't run VBoxService because the
-	#  32bit VBoxService won't work with the 64bit kernel modules, but the 64bit
-	#  VBoxService won't work with our 32bit userspace; good times)
+	for line in $(VBoxControl --nologo sharedfolder list -automount | tail -n+3 | cut  -d " " -f 3)
+	do
+	    try_mount_share "$line";
+	done
 fi


tianon:
I like where this is going, but I am curious whether at this point it might not just make more sense to use `VBoxService --only-automount` like the comment suggests, avoiding our own custom shell logic entirely; I don't know what the full implications of that (or even this) might be, so I want to be very cautious in proceeding with anything like this.

cc @nathanleclaire 

For example, direct mounts like `/home`, `/etc`, or worse `/usr` would break the VM OS in pretty horrific ways.

nathanleclaire:
cc @dgageot @jeanlaurent as well

nathanleclaire:
Isn't the userspace also 64bit now?

tianon:
Yes, it is (which is what makes any of this possible). :smile:

u-foka:
@tianon: I think the problem with --only-automount is that it will not do the uid/gid mapping for us :/

nathanleclaire:
Which OS are you testing on?  Have you tried OSX, Windows, _and_ Linux?  Pathing issues seem likely to me.

u-foka:
I've only tested it on osx.
It should work with the default windows share name: "c/Users", and now I've tested that it can even mount totally fu**ed up share names like c:\Users if the user requests it using virtualbox-share-folder from https://github.com/docker/machine/pull/2100

dgageot:
Hi @u-foka and @tianon. What's the update on this PR? There's a docker-machine PR (https://github.com/docker/machine/pull/2100) waiting for this to be merged.
Is it something that could be merged as is? Does it need improvements from @u-foka?

nathanleclaire:
Main thing IMO is to verify that default functionality still works OK on Linux and Windows before merging this in.

djmaze:
Can confirm this works as expected on Linux: When creating a shared folder named `/home/user`, that folder is automatically mounted at `/home/user` in the VM.

u-foka:
Related PR in docker/machine is now: https://github.com/docker/machine/pull/3798

rootfs/rootfs/etc/rc.d/vbox
```diff
@@ -33,15 +33,8 @@ if modprobe vboxguest &> /dev/null && modprobe vboxsf &> /dev/null; then
...
-		|| try_mount_share /Users \
-		|| try_mount_share /c/Users 'c/Users' \
-		|| try_mount_share /c/Users \
-		|| try_mount_share /c/Users 'c:/Users' \
-		|| true
-	# TODO replace this whole hacky bit with VBoxService --only-automount
-	# (the problem with that being that we can't run VBoxService because the
-	#  32bit VBoxService won't work with the 64bit kernel modules, but the 64bit
-	#  VBoxService won't work with our 32bit userspace; good times)
+	for line in $(VBoxControl --nologo sharedfolder list -automount | tail -n+3 | cut  -d " " -f 3)
```
yosifkit: I feel like the cut on spaces is an incomplete solution.  What happens when a user decides to use a share with a space in the path?  ie: `c:\Program Files`

SergeC:
Can it be merged into upcoming v1.13.0?

PR #1089, opened 2015-11-04
grammar and formatting

aranaur:
Fixed formatting to style matched steps one and two of delete/create VM.

Switched "changes" --> "change" as only one change is mentioned.


Moghedrin:
LGTM

PR #1079, opened 2015-10-16
Compile kernel modules for ipvsadm

glinton:
Include kernel modules required for `ipvsadm`

PR #1104, opened 2015-12-16
Adding ipvsadmin binary

stelcheck:
Related to https://github.com/boot2docker/boot2docker/pull/1079 

I thought it didn't really make sense to have the IPVS modules without ipvsadm installed... I guess installing it from Debian and then dropping it into the image isn't clean, but since Tiny Core Linux isn't maintaining a package for ipvsadm, I thought that might just be fine.

Please let me know if that's too ugly of a hack to merge.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,6 +17,7 @@ RUN apt-get update && apt-get -y install  unzip \
                         isolinux \
                         automake \
                         pkg-config \
+                        ipvsadm \
                         p7zip-full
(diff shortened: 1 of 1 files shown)


tianon:
Indeed -- copying this binary from Debian directly as-is is quite a bit hairier than I'm comfortable with.  Can't this be run from inside a container instead? (using `--device` to give it access to the appropriate device nodes)

PR #1116, opened 2016-01-15
Fix typo in /etc/rc.d/sshd

### LINKED ISSUE #1112: possible bad file name in rootfs/rootfs/etc/rc.d/sshd script
I am not an expert, but I just noticed in boot2docker.log file an error caused by bad file name in the following code:

    if [ ! -f /usr/local/etc/ssh/ssh_config ]; then
    	    mv /usr/local/etc/ssh/ssh_config.example /usr/local/etc/ssh/ssh_config

ssh script wants to rename `ssh_config.example` file, while in the iso file these is no such file, instead there is a file named `ssh_config_example` **with underscore** between ssh_config and example.

This is in Docker 1.9.1 and boot2docker 1.9.1 build cef800b

AkihiroSuda:
Fix #1112


tianon:
Nice, thanks!   LGTM

PR #1118, opened 2016-01-17
Enable usbip (usb over ip) server/client kernel module

dweinstein:
This enables pushing USB packets over IP to allow remote machines direct access to USB devices. This change enables both the "server" component and the "client" component. 

This enables sharing a usb device hooked up to the docker host or vice versa.


dweinstein:
oops got a little trigger happy. I want to test this more before the merge since we're only on 4.1.15. Closing for now.

dweinstein:
..in progress..

from a docker container, listing the remote apple device hooked up to another linux box:

```
root@0585a670215a:~# ./usbip list -r 10.0.1.10
Exportable USB devices
======================
 - 10.0.1.10
      2-1.1: Apple, Inc. : iPhone5/5C/5S (05ac:12a8)
           : /sys/devices/pci0000:00/0000:00:1d.0/usb2/2-1/2-1.1
           : (Defined at Interface level) (00/00/00)

```

RomanSaveljev:
I have been having this module in a private boot2docker fork for around half a year now. It runs decently for me.

:+1: for integrating it

tianon:
Do you happen to have the ISO size difference before/after numbers handy?

tianon:
(I'm guessing it's negligible in which case :+1:, but I'd like to double check to be sure.)

dweinstein:
about 1 MB bigger when I compare to the bootdocker.iso I got from docker toolbox, which surprised me a little tbh it was that much

dweinstein:
```± wc -c ~/.docker/machine/machines/default/boot2docker.iso ~/.docker/machine/machines/usbip/boot2docker.iso
 31457280 /Users/user/.docker/machine/machines/default/boot2docker.iso
 32505856 /Users/user/.docker/machine/machines/usbip/boot2docker.iso
```

tianon:
Nice!

I think that increase is 100% Docker itself; the rc1 I built last week (https://github.com/tianon/boot2docker-legacy/releases/tag/v1.10.0-rc1):
```
32505856 boot2docker.iso
```

(Same as your "new" value, so we're golden.)

tianon:
LGTM

PR #1119, opened 2016-01-18
add DOCKER_ENV file to specify download subdomain

ChristianKniep:
Hey there,

I made a small adjustment to be able to create b2d with 1.10.0-rc1, which is downloaded from `test.docker.com` instead of `get.docker.com`.

Cheers
Christian


tianon:
Hmm, I'm not very fond of this as-is, given how easy it is to adjust after the fact and how this further complicates an already very involved `Dockerfile`.  For reference, here's how I build the official RCs (https://github.com/tianon/boot2docker-legacy/releases):
```Dockerfile
FROM boot2docker/boot2docker

ENV TEST_DOCKER_VERSION 1.10.0-rc1

RUN curl -fsSL https://test.docker.com/builds/Linux/x86_64/docker-$TEST_DOCKER_VERSION -o $ROOTFS/usr/local/bin/docker \
	&& chmod +x $ROOTFS/usr/local/bin/docker \
	&& $ROOTFS/usr/local/bin/docker -v

RUN $ROOTFS/usr/local/bin/docker -v | sed -r 's/.* version ([^ ,]+).*/\1/' > $ROOTFS/etc/version \
	&& cp -v $ROOTFS/etc/version /tmp/iso/version

RUN { echo; echo "  WARNING: this is a test.docker.com build, not a release."; echo; } >> $ROOTFS/etc/motd

RUN /make_iso.sh
```

(saved as `Dockerfile.test`)

```console
$ docker build -t boot2docker/boot2docker --pull .
$ docker build -t boot2docker/boot2docker:test -f Dockerfile.test .
$ docker run --rm boot2docker/boot2docker:test > boot2docker.iso
```

ChristianKniep:
I see... that make sense... :) I looked at it but somehow got it wrong.

PR #1121, opened 2016-01-18
Persist /etc/docker directory

dhiltgen:
Starting in docker 1.10, a persistent configuration file can be stored
(/etc/docker/daemon.json).  This change makes sure that we store
/etc/docker on the mounted filesystem so it will survive reboots.

### DIFF
--- a/rootfs/rootfs/usr/local/etc/init.d/docker
+++ b/rootfs/rootfs/usr/local/etc/init.d/docker
@@ -28,6 +28,11 @@ test -f '/var/lib/boot2docker/profile' && . '/var/lib/boot2docker/profile'
 export PATH=${PATH}:/usr/local/sbin
 
 start() {
+    if [ ! -e "/etc/docker" ]; then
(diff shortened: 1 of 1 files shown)


thaJeztah:
ping @tianon @SvenDowideit could you please review this? We'd like to have this in 1.10 if possible

cpuguy83:
Do you think it would be better for docker to start with `--config-file=/var/lib/boot2docker/daemon.json` instead?

dhiltgen:
closes https://github.com/docker/docker/issues/19454

dhiltgen:
@cpuguy83 I would really prefer to stick with the standard path, since other tools/automation are likely to be built around that default path assumption.

tianon:
Nice and simple, and makes sure hard-coded assumptions continue to work properly. :+1:

LGTM

dhiltgen:
Thanks all!

thaJeztah:
Thanks @tianon!

PR #1132, opened 2016-02-05
Fix /etc/init.d/docker restart

huguesb:
Docker daemon takes forever and a half to exit (about 10s on a
fresh boot2docker 1.10 VM). Attempting to restart before it is
stopped just fails miserably: start aborts and stop eventually
completes.

Arguably start and stop should only return once the daemon is
actually started/stopped but that might impose undue delay on
callers that only care about an eventual state change and are
unmoved by the prospect of race conditions.

### DIFF
--- a/rootfs/rootfs/usr/local/etc/init.d/docker
+++ b/rootfs/rootfs/usr/local/etc/init.d/docker
@@ -114,10 +114,15 @@ stop() {
 
 restart() {
     if check; then
-        stop && sleep 1 && start
-    else
-        start
+        stop
+        i=30
+        while check ; do
+            sleep 1
+            i=$(expr $i - 1)
+            [ "$i" -gt 0 ] || { echo "Failed to stop Docker dameon" ; exit 1 ; }
+        done
     fi
+    start
 }
 
 check() {


AkihiroSuda:
+1, can be simplified like this?

```bash
restart() {
    if check; then
        stop
        while check; do
	        sleep 1
        done
    fi
    start
}
```

huguesb:
@AkihiroSuda done

rootfs/rootfs/usr/local/etc/init.d/docker
```diff
@@ -114,10 +114,12 @@ stop() {
 
 restart() {
     if check; then
-        stop && sleep 1 && start
-    else
-        start
+        stop
+        while check ; do
+            sleep 1
+        done
```
tianon: If the Docker daemon somehow doesn't get signaled properly during `stop`, this will essentially be an infinite loop -- do you think it warrants adding some simple "max tries" logic to give us a timeout?

huguesb:
@tianon Done

30s seems like a reasonable timeout given that I've seen the daemon take 10-15s to stop.

tianon:
Thanks :+1:

LGTM

Issue #992, opened 2015-07-15
Allow mkfs.ext4 -i in etc/rc.d/automount


clkao:
With overlay storage driver it can easily run out of inode with the default settings.  I think coreos made a change a while back about this.

We should cosnider using `-i 1024` by default or allow inode size to be configured somehow.

lox:
This is a huge problem for our team of 30 now using docker-machine. We run into this on a daily basis and need to destroy and recreate our VMs. 

Are there any temporary workarounds that will let us tune the ratio of inodes?

https://github.com/docker/docker/issues/9755
https://github.com/coreos/bugs/issues/264

PR #1135, opened 2016-02-09
Use a higher ratio of inodes to address inode exhaustion in overlayfs

lox:
We've been encountering regular inode exhaustion with boot2docker/docker-machine. It seems like this is a problem that is likely to be affecting more and more people with the gradual convergence on overlayfs. 

I'd hoped to be able to pass a parameter somehow to the `automount` script, I guess via a kernel boot parameter? That would address #992. I couldn't figure out a way to do that, any suggestions?

### DIFF
--- a/rootfs/rootfs/etc/rc.d/automount
+++ b/rootfs/rootfs/etc/rc.d/automount
@@ -49,7 +49,7 @@ if [ ! -n "$BOOT2DOCKER_DATA" ]; then
                 sleep 0.5
             done
             BOOT2DOCKER_DATA=`echo "${UNPARTITIONED_HD}1"`
-            mkfs.ext4 -L $LABEL $BOOT2DOCKER_DATA
+            mkfs.ext4 -i 8192 -L $LABEL $BOOT2DOCKER_DATA
             swapon "${UNPARTITIONED_HD}2"
         fi
 
@@ -74,7 +74,7 @@ if [ ! -n "$BOOT2DOCKER_DATA" ]; then
                     # Add the data partition
                     (echo n; echo p; echo 1; echo ; echo ; echo w) | fdisk $UNPARTITIONED_HD
                     BOOT2DOCKER_DATA=`echo "${UNPARTITIONED_HD}1"`
-                    mkfs.ext4 -L $LABEL $BOOT2DOCKER_DATA
+                    mkfs.ext4 -i 8192 -L $LABEL $BOOT2DOCKER_DATA
                     swapon "${UNPARTITIONED_HD}2"
                 else
                     echo "Disk unpartitioned but something is there... not doing anything"


lox:
Looks like my whitespace settings have caused additional diffs. Will correct shortly.

rootfs/rootfs/etc/rc.d/automount
```diff
@@ -49,7 +49,7 @@ if [ ! -n "$BOOT2DOCKER_DATA" ]; then
                 sleep 0.5
             done
             BOOT2DOCKER_DATA=`echo "${UNPARTITIONED_HD}1"`
-            mkfs.ext4 -L $LABEL $BOOT2DOCKER_DATA
+            mkfs.ext4 -T news -L $LABEL $BOOT2DOCKER_DATA
```
tianon: Does our `mkfs.ext4` implementation include the `-T` flag?  Does it include `/etc/mke2fs.conf` to define what `news` means?  Would it be safer to instead just use `-i` directly and figure out an OK value?  I've had good luck with `-i 8096` myself generally, but did recently have to reformat my home machine to `-i 6072` (but granted that I do a lot of work on the Docker Official Images, so I have thousands++ of images/layers on a constant, cycling basis so I think I'm a bit of an edge case here I think).

lox: Yup, I think 4096 or 8192 would be sensible. I'll update the PR.

tianon: :metal: be sure to give me a prod when it's ready!

nathanleclaire:
Cool!  Curious to see where this goes

lox:
I've updated this based on feedback, presently building a local test image. Are there any other places that `mkfs.ext4` is called that need updating?

tianon:
Not that I know of! :metal:

LGTM (but I'd like to have @nathanleclaire's :+1: on this too before merge)

nathanleclaire:
I'll bake ISO and smash on it a bit.

nathanleclaire:
Seems to be handling `docker-stress` (`docker-stress --containers 1000 --concurrent 10`) pretty OK.

nathanleclaire:
Ran `docker-stress` quite a bit and also did a little thing like:

```
$ for i in $(seq 0 1000); do docker run -d -P nginx; done
```

My VM ran out of memory before it ran out of inodes.

LGTM

nathanleclaire:
@yosifkit You might want to consider this one for the release tomorrow.

yosifkit:
@tianon and @nathanleclaire agree, so LGTM

nathanleclaire:
:+1:

lox:
Nice, thanks all!

PR #1137, opened 2016-02-10
Add proc_events netlink API kernel support

ashb:
This allows a process with CAP_NET_ADMIN to ask the kernel to notify
processes via a netlink socket when ever a process or thread/task is created.
This is needed for some monitoring and security tools to run.

It was added to debian mainline in 2012 in this bug
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=588200

I had to switch connector from a module to built-in otherwise during
`make oldconfig` it would disable PROC_EVENTS again.


tianon:
Reasonable, thanks! LGTM :+1:

PR #1152, opened 2016-03-22
Less permissive file modes for /Users mount

embray:
When mounting C:\Users and the like, at least on Windows (I'm not sure if this is the same if the host OS is OSX in which case this might need adjusting) all files are given mode `777`.  This is a problem for a few reasons that I listed in sagemath/docker-images#19

Setting `600` for files seems to make sense, since for directories under `/Users`/`C:\Users` most files should be private anyways.  I'm split on whether files should be executable or not, but ultimately going with not to be on the safe side.


tianon:
I'm hesitant on this one given that it's technically a breaking change. :disappointed:

(Bind-mounting this directory into various containers running as various UIDs is likely to cause all sorts of trouble, for example.)

PR #1161, opened 2016-04-19
Add dummy link to libtirpc to fix vmhgfs-fuse

### LINKED ISSUE #1157: usr/local/bin/vmhgfs-fuse: error while loading shared libraries: libtirpc.so.1
```
# Install build dependencies for VMware Tools

        libtirpc1 \

/usr/local/bin/vmhgfs-fuse: error while loading shared libraries: libtirpc.so.1: cannot open shared object file: No such file or directory
/usr/local/bin/vmtoolsd: error while loading shared libraries: libtirpc.so.1: cannot open shared object file: No such file or directory
/usr/local/bin/vmhgfs-fuse: error while loading shared libraries: libtirpc.so.1: cannot open shared object file: No such file or directory
```

TCL 7 supplies

```
/usr/local/lib/

libtirpc.so -> libtirpc.so.3.0.0
lrwxrwxrwx    1 root     root            17 Apr 15 00:37 libtirpc.so.3 -> libtirpc.so.3.0.0
-rwxr-xr-x    1 root     root        137064 Apr 10 14:38 libtirpc.so.3.0.0
```

Therefore I am unable to mount volumes via vmware fusion like I once did. This seems to have happen ed after the docker 1.11.0 update. Previously this was working fine.

frapposelli:
Fixes #1157

Signed-off-by: Fabio Rapposelli <fabio@vmware.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -209,6 +209,11 @@ RUN curl -fL -o /tmp/${LIBDNET}.zip https://github.com/dugsong/libdnet/archive/$
 RUN cd $ROOTFS && cd usr/local/lib && ln -s libdnet.1 libdumbnet.so.1 &&\
     cd $ROOTFS && ln -s lib lib64
 
+# TCL 7 doesn't ship with libtirpc.so.1 Dummy it up so the VMware tools work
+# again, taken from:
+# https://github.com/boot2docker/boot2docker/issues/1157#issuecomment-211647607
+RUN cd $ROOTFS/usr/local/lib && ln -s libtirpc.so libtirpc.so.1 && readlink -f libtirpc.so.1
+
 # Download and build Parallels Tools
 ENV PRL_MAJOR 11
 ENV PRL_VERSION 11.1.0


Dockerfile
```diff
@@ -209,6 +209,11 @@ RUN curl -fL -o /tmp/${LIBDNET}.zip https://github.com/dugsong/libdnet/archive/$
 RUN cd $ROOTFS && cd usr/local/lib && ln -s libdnet.1 libdumbnet.so.1 &&\
     cd $ROOTFS && ln -s lib lib64
 
+# TCL 7 doesn't ship with libtirpc.so.1 Dummy it up so the VMware tools work
+# again, taken from:
+# https://github.com/boot2docker/boot2docker/issues/1157#issuecomment-211647607
+RUN cd $ROOTFS/usr/local/lib && ln -s libtirpc.so.3.0.0 libtirpc.so.1 && ldconfig
```
tianon: Would it also work to just `ln -s libtirpc.so libtirpc.so.1` to insulate ourselves a little bit from future changes?  (`ldconfig` shouldn't be necessary either -- that'd update the LD cache for the build environment, which shouldn't be looking inside `$ROOTFS` anyhow)

tianon: Maybe follow it up with a `&& readlink -f libtirpc.so.1` which will fail if our symlink ever points to something invalid, too?

frapposelli:
@tianon Looks better now 😄

tianon:
LGTM :+1:

PR #1162, opened 2016-04-20
Reorganize the Dockerfile a bit (for caching, but also for testability)

tianon:
This allows us to run binaries like "vmhgfs-fuse" and verify that at the very least all the ".so" files they need are discoverable.  (Which is really important for helping us avoid issues like #1157 in future updates.)

cc @nathanleclaire @frapposelli

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -60,25 +60,7 @@ RUN jobs=$(nproc); \
 
 # The post kernel build process
 
-ENV ROOTFS          /rootfs
-ENV TCL_REPO_BASE   http://tinycorelinux.net/7.x/x86_64
-# Note that the ncurses is here explicitly so that top continues to work
-ENV TCZ_DEPS        iptables \
-                    iproute2 \
-                    openssh openssl \
-                    tar \
-                    gcc_libs \
-                    ncurses \
-                    acpid \
-                    xz liblzma \
-                    git expat2 libiconv libidn libgpg-error libgcrypt libssh2 \
-                    nfs-utils tcp_wrappers portmap rpcbind libtirpc \
-                    rsync attr acl \
-                    curl ntpclient \
-                    procps glib2 libtirpc libffi fuse pcre \
-                    udev-lib udev-extra \
-                    liblvm2 \
-                    parted
+ENV ROOTFS /rootfs
 
 # Make the ROOTFS
 RUN mkdir -p $ROOTFS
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -128,24 +110,52 @@ RUN cd /linux-kernel && \
...
 # Download the rootfs, don't unpack it though:
 RUN curl -fL -o /tcl_rootfs.gz $TCL_REPO_BASE/release/distribution_files/rootfs64.gz
 
 # Install the TCZ dependencies
-RUN for dep in $TCZ_DEPS; do \
-    echo "Download $TCL_REPO_BASE/tcz/$dep.tcz" &&\
-        curl -fL -o /tmp/$dep.tcz $TCL_REPO_BASE/tcz/$dep.tcz && \
-        unsquashfs -f -d $ROOTFS /tmp/$dep.tcz && \
-        rm -f /tmp/$dep.tcz ;\
+RUN set -ex && \
```
tianon: Adding `set -e` here helped me find a number of the `TCZ_DEPS` that don't exist anymore (and don't seem to be necessary either), since they 404 and fail now.  I think `libiconv` and `libidn` were the two most notable examples.

tianon:
Since we're resolving deps-of-`TCZ_DEPS` ourselves, I wonder if it'd be prudent to also include a "`chroot` test" for some of the utilities we install via that method too (like `git --version`, etc).  I also wonder if we could instead utilize TCL's package manager more directly and get dep resolution somewhat more automatically (because I _have_ to believe it manages "dependency" information for each package :sweat_smile:).

tianon:
(I should add that I've tested both that this builds successfully, and that the resulting ISO still works and Docker still runs containers within it.)

Dockerfile
```diff
@@ -324,4 +345,4 @@ COPY rootfs/make_iso.sh /
 
 RUN /make_iso.sh
 
-CMD ["cat", "boot2docker.iso"]
+CMD ["sh", "-c", "[ -t 1 ] && exec bash || exec cat boot2docker.iso"]
```
tianon: This one is so that our existing flow of `docker run --rm boot2docker/boot2docker > boot2docker.iso` is still unchanged, but doing `docker run -it --rm boot2docker/boot2docker` doesn't start outputting an ISO...  :smile:  (NOT AGAIN :sob:  *slams the mute keystroke*)

frapposelli: so much this!! 😄 

gif-keyboard-18384011143327871983

Dockerfile
nathanleclaire: Nice

Dockerfile
nathanleclaire: I know this pain

PR #1163, opened 2016-04-22
Allow spaces in certificate file names

davideaster:
Although it's conventional for cert file names to use underscores and not spaces, this fix allows `/etc/rc.d/install-ca-certs` to work if `/var/lib/boot2docker/certs` names do contain spaces.


yosifkit:
LGTM

Issue #735, opened 2015-02-12
how to add self-signed certificate to boot2docker


hordemark:
I have a self-signed certificate named ca.pem, how to add self-signed certificate to boot2docker.
There are a lot of solutions on the net, but none works for me.
Is there anyone succeed on your environment? Thanks!

hordemark:
Resoved

https://github.com/boot2docker/boot2docker/issues/347#issuecomment-70950789

PR #1167, opened 2016-05-27
Also process certificates with the .crt extension.

gotgenes:
The [official Docker docs on creating a self-signed certificate](https://docs.docker.com/registry/insecure/#using-self-signed-certificates) use a `.crt` extension for the public PEM certificate. To help users that are following those documents, it would help to also accept that extension in this script, as well.

### DIFF
--- a/rootfs/rootfs/etc/rc.d/install-ca-certs
+++ b/rootfs/rootfs/etc/rc.d/install-ca-certs
@@ -5,7 +5,7 @@ BOOT2DOCKER_CERTS_DIR=/var/lib/boot2docker/certs
 CERTS_DIR=/etc/ssl/certs
 CAFILE=${CERTS_DIR}/ca-certificates.crt
 
-for cert_file in "${BOOT2DOCKER_CERTS_DIR}"/*.pem; do
+for cert_file in "${BOOT2DOCKER_CERTS_DIR}"/*.pem "${BOOT2DOCKER_CERTS_DIR}"/*.crt; do
 	cert=`basename "$cert_file"`
 	echo "installing $cert"
 


tianon:
Given https://github.com/boot2docker/boot2docker/blob/d63b52eefd94c793ec7ecb3bc198eb425fd78ed9/rootfs/rootfs/usr/local/etc/init.d/docker#L31-L35 (persistent `/etc/docker`), wouldn't it be better to follow the official documentation and use `/etc/docker/certs.d` for this now?

gotgenes:
> Given https://github.com/boot2docker/boot2docker/blob/d63b52eefd94c793ec7ecb3bc198eb425fd78ed9/rootfs/rootfs/usr/local/etc/init.d/docker#L31-L35 (persistent `/etc/docker`), wouldn't it be better to follow the official documentation and use `/etc/docker/certs.d` for this now?

Possibly? Documentation in the main `README.md` would go a long way to help here, as well as a definitive answer in the FAQ of "How do I add a self-signed certificate to boot2docker?". `/var/lib/boot2docker/etc/docker` was at no point mentioned in #347. That change was introduced in January, yet boot2docker users are still searching for answers as the comments in #347, and the fact that it is still an open issue, show. #347 is the second hit on Google for me with the search "boot2docker self signed certificate" (the top hit being #735, which points to #347).

gotgenes:
@tianon Upon further review of the [Docker self-signed certificates documentation](https://docs.docker.com/registry/insecure/#docker-still-complains-about-the-certificate-when-using-authentication), placing certificates in `/etc/docker/certs.d` may be insufficient:

> When using authentication, some versions of docker also require you to trust the certificate at the OS level.

This latter action (making the OS trust the certificates) appears to be the role of [install-ca-certs](https://github.com/boot2docker/boot2docker/blob/master/rootfs/rootfs/etc/rc.d/install-ca-certs), which processes the certificates in `/var/lib/boot2docker/certs`.

bbodenmiller:
While the new method of `/etc/docker/certs.d` is good I don't think you can remove the old method as many are using it so I think it is worth merging this. Also updating `README.md` to state that `.pem` and `.crt` are allowed.

I've also opened a PR at https://github.com/boot2docker/boot2docker/pull/1195 to document the new method of `/etc/docker/certs.d`.

yosifkit:
LGTM, cc @tianon.

tianon:
Fair enough, LGTM

gotgenes:
Thanks, all!

PR #1170, opened 2016-06-10
Build IPVS directly into kernel for 1.12

dhiltgen:
As modules, these aren't being loaded and break
some of the 1.12 networking features.


yosifkit:
`ip_vs` is also marked as a module in [Debian](https://sources.debian.net/src/linux/4.6.1-1/debian/config/config/#L6413).  As far as I can tell, this depends on https://github.com/docker/docker/pull/23022.  The normal docker way is to do modprobe before using any kernel feature that can be a module, so I think this is a bug in the docker branch that you are using.

mrjana:
@dhiltgen After the latest libnetwork vendor in docker/docker the module gets auto loaded the first time ipvs is needed.

tianon:
Nice, thanks @mrjana!

dhiltgen:
Sounds like this is no longer needed.  Thanks!

PR #1173, opened 2016-06-16
Changed kernel parameters to support cattle overlay networking.

jd-daniels:
Maybe no one else is interested in this. But currently, cattle orchestration can't work with boot2docker because the overlay network requires some things for the tunneling to work. Changing three kernel params for XFRM seems to fix this.

I compared the kernel config requirements from here:
https://wiki.strongswan.org/projects/strongswan/wiki/KernelModules


tianon:
Given that Debian enables them (https://sources.debian.net/src/linux/4.6.1-1/debian/config/config/#L6086), I'm +1.  Can we use `=m` instead of `=y` though?

jd-daniels:
Changed to '=m'. Modules make sense for this.

tianon:
Nice, ISO size is unchanged, and since they're modules, they should have no effect on any existing functionality by default. :+1:

LGTM

jd-daniels:
Tested the new iso via rancher and now boot2docker works fine with cattle networking out of the box. Thanks! :+1:

tianon:
:metal:

PR #1203, opened 2016-09-24
enable SCTP module in kernel config

colrack:
This enables SCTP support in kernel


yosifkit:
What is the specific need for this?  Mostly I would like to have a documented reason for enabling it.

In any case, it seems fair to enable [Stream_Control_Transmission_Protocol](https://en.wikipedia.org/wiki/Stream_Control_Transmission_Protocol), since these two flags are also enabled in debian: https://sources.debian.net/src/linux/4.7.5-1/debian/config/config/#L6278 and https://sources.debian.net/src/linux/4.7.5-1/debian/config/config/#L6576.

cc @tianon

tianon:
Agreed -- I'm +1 on adding it, but would love to hear a bit more about the specific use case that's prompted this addition (even if only for posterity). :+1:

colrack:
As you may know SCTP is a transport level protocol like TCP and UDP.
All major distribution have SCTP enabled.
I use SCTP everyday during development, coding both in C and Erlang.
To use SCTP I need the host to have SCTP enabled in kernel.
This is my specific need, and since I use boot2docker I would see this harmless change to go mainstream and forget about building my boot2docker iso every time there is a release or use another vm (with another distro) as docker host.

This PR is related and enable SCTP in official erlang image:
https://github.com/c0b/docker-erlang-otp/pull/16

tianon:
Excellent, thanks @colrack!

PR #1204, opened 2016-09-28
Apply workaround for apparent TCL "openssl" changes (especially with regards to "/etc/ssl")

tianon:


tianon:
For some context, it would appear that TCL's `openssl` package has changed out from under us and now requires a little bit of additional setup after it's extracted (which I'm guessing TCL's package manager would've handled automatically for us, but we get to apply manually).

This fixes the following when attempting to pull images on recent builds:
```
docker: Error while pulling image: Get https://index.docker.io/v1/repositories/library/alpine/images: x509: failed to load system roots and no roots provided.
```

tianon:
Updated the comments to be a bit more clear and to include some testing to ensure it's actually doing what it's supposed to.

tianon:
(Currently doing another test build to ensure it's working 100% properly before I merge and rebuild official v1.12.2-rc1 ISOs.)

tianon:
Success!
```console
docker@boot2docker:~$ docker run -it --rm alpine sh
Unable to find image 'alpine:latest' locally
latest: Pulling from library/alpine
c0cb142e4345: Pull complete 
Digest: sha256:ca7b185775966003d38ccbd9bba822fb570766e4bb69292ac23490f36f8a742e
Status: Downloaded newer image for alpine:latest
/ # 
```

nathanleclaire:
Blergh, thanks for taking care of this.

Issue #1209, opened 2016-10-18
version v1.12.2 failed to build on docker hub


cglewis:
https://hub.docker.com/r/boot2docker/boot2docker/builds/bpdxyrkgcgt72aijqo2uwcf/

```
Build failed: The command '/bin/sh -c curl -fL -o /tcl_rootfs.gz $TCL_REPO_BASE/release/distribution_files/rootfs64.gz' returned a non-zero code: 7
```

latest available is still only 1.12.1 build 4b170dc

cglewis:
any update on this?

yosifkit:
The build has been triggered 3, 2, and 1 days ago, both manually and via master updates.  I manually triggered it once again to see if it will finish this time.

cglewis:
looks like it just keeps error-ing out.  are there alternatives we can do in the mean time? it seems this happens quite often, maybe a longer term solution is in order...

cglewis:
looks like this issue might be the cause: https://github.com/docker/hub-feedback/issues/811

yosifkit:
We are going to [start pushing them](https://github.com/tianon/jenkins-groovy/commit/1709f3a930ffabb39742b04ae759b588934c61e7) as part of the process that builds the iso, rather than relying on automated builds.  As a bonus we'll have tags like `boot2docker/boot2docker:1.12.2`.

cglewis:
@yosifkit cool. as an aside, I think the PR I opened (referenced above) will solve the problem that the automated build was experiencing. 

sadly this project seems like a long lost pet project at this point and doesn't get as much attention as the ones under the `docker` org.  is this project looking for maintainers?  i'd really like to keep it alive, or even if there's a next evolution of it (i think there was an attempt to do something debian based at one point)...

yosifkit:
This should be fixed by https://github.com/boot2docker/boot2docker/pull/1214.

PR #1214, opened 2016-10-24
workaround instead of adding files to /

cglewis:
- addresses issues where build on the hub fails, such as #1209
- uses workaround based on docker/hub-feedback#811


tianon:
Thanks @cglewis :heart:

I don't understand why it fails, but I do trust that it's better this way. :+1:

PR #1221, opened 2016-11-19
Don't attempt getty more than once

nathanleclaire:
@tianon 

Closes: https://github.com/docker/machine/issues/3711

Signed-off-by: Nathan LeClaire <nathan.leclaire@gmail.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -329,8 +329,8 @@ RUN mv $ROOTFS/shutdown.sh $ROOTFS/opt/shutdown.sh && \
 RUN echo "#!/bin/sh" > $ROOTFS/usr/local/bin/autologin && \
 	echo "/bin/login -f docker" >> $ROOTFS/usr/local/bin/autologin && \
 	chmod 755 $ROOTFS/usr/local/bin/autologin && \
-	echo 'ttyS0:2345:respawn:/sbin/getty -l /usr/local/bin/autologin 9600 ttyS0 vt100' >> $ROOTFS/etc/inittab && \
-	echo 'ttyS1:2345:respawn:/sbin/getty -l /usr/local/bin/autologin 9600 ttyS1 vt100' >> $ROOTFS/etc/inittab
+	echo 'ttyS0:2345:once:/sbin/getty -l /usr/local/bin/autologin 9600 ttyS0 vt100' >> $ROOTFS/etc/inittab && \
+	echo 'ttyS1:2345:once:/sbin/getty -l /usr/local/bin/autologin 9600 ttyS1 vt100' >> $ROOTFS/etc/inittab
 
 # fix "su -"
 RUN echo root > $ROOTFS/etc/sysconfig/superuser


jonathanperret:
Hi @nathanleclaire, I'm afraid this kind of breaks the functionality of being able to use serial ports as a console. I just tested that with `getty` set to be started only `once`, one can get a to a shell by connecting to the serial port, but once that shell exits, nothing listens on the port anymore (`getty` does not hang around). Is that intentional?

tianon:
@jonathanperret hmm, good point :disappointed:

I wonder if there's some way we can mimic systemd's generator behavior here and do some runtime detection to determine whether the given tty exists before launching `getty` repeatedly on it? :confused:

simonvanderveldt:
What might work is having the entry in `inittab` execute a script that checks if the relevant `/dev/tty` device exists and then (re)starts `getty` for the tty and if it doesn't it just sleeps indefinitely?

So something like this in inittab:
```
null::respawn:/path/to/script /dev/ttysomething
```

Not sure of how "ugly"/OK this is though :x

tianon:
@simonvanderveldt seems like a pretty sane approach to me!

Any idea if there's an easy way for us to determine whether a given TTY is actually connected?
```console
docker@boot2docker:~$ ls -l /dev/ttyS*
crw-------    1 docker   staff       4,  64 Jan  4 21:24 /dev/ttyS0
crw--w----    1 root     root        4,  65 Jun 17  2006 /dev/ttyS1
crw-rw----    1 root     staff       4,  66 Jun 17  2006 /dev/ttyS2
crw-rw----    1 root     staff       4,  67 Jun 17  2006 /dev/ttyS3
```

Maybe there's something in the code systemd uses for doing so we could mimic?

tianon:
https://github.com/systemd/systemd/blob/6a47fd894d601f7e8e88dec4cb35dfb7d7c15eff/src/getty-generator/getty-generator.c#L94-L116 seems to be the relevant code:
```c
        /* Some TTYs are weird and have been enumerated but don't work
         * when you try to use them, such as classic ttyS0 and
         * friends. Let's check that and open the device and run
         * isatty() on it. */
```

tianon:
:smiling_imp: 
```console
root@boot2docker:~# exec 42</dev/ttyS1 && { test -t 42 && echo yes || echo no; }
 && exec 42<&-
no
root@boot2docker:~# exec 42</dev/ttyS0 && { test -t 42 && echo yes || echo no; }
 && exec 42<&-
yes
```

simonvanderveldt:
@tianon Not really sure what the above voodoo is :P
(looks pretty interesting though, will try it out when I can)

Also, I'm not sure if my initial idea of sleeping indefinitely will work for the situation described by @jonathanperret because that would still mean no restart of getty when initially there's no serial console connected.

After some more pondering I came to the conclusion that it would be perfect if udev could do this using some udev rules, that way it would be properly dynamic.
I have no clue though if there's a way to detect from udev when a serial port gets connected. AFAIK there's some signalling in the port itself, but I'm not sure if that changes a udev attribute.
Maybe something like this will work? Will have to do some trial and error to see if this works
```
SUBSYSTEM=="tty", ACTION=="add", ATTRS{type}=="0", RUN+="/path/to/script"
```

jonathanperret:
The problematic restarts do not occur because nothing is connected to the serial port — `getty` is perfectly happy listening on an _unconnected_ serial port.

The problem exists because those serial ports do not even exist in the VM! Just enabling them in the VM config would be enough to make `getty` happy, but for some reason `machine` configures the VM to not have serial ports enabled.

simonvanderveldt:
@jonathanperret sorry, then I guess I misunderstood. So we can start getty as longs as the port exists, independent of if something's connected to it.

But this:
> I just tested that with getty set to be started only once, one can get a to a shell by connecting to the serial port, but once that shell exits, nothing listens on the port anymore (getty does not hang around). Is that intentional?

does mean that once disconnected `getty` exits and will need to be restarted again (which the inittab + script solution would do, so should be fine), correct?

jonathanperret:
Yes, `getty` opens the serial port, waits for a connection, handles it and then exits. It's `init`'s job to restart it (which is why `once` doesn't cut it).

Note that the presence or absence of a specific serial port is highly unlikely to change while the VM is started (VirtualBox does not let you enable serial ports while a VM is running), so that's not really a scenario that you need to worry about. Sleeping indefinitely would probably work. But that looks like a complicated workaround when `docker-machine` could simply enable the serial ports instead…

simonvanderveldt:
@nathanleclaire on the other hand the same issue as described in https://github.com/docker/machine/issues/3711 would happen for every host (VM or bare metal) that doesn't enable `ttyS0` and `ttyS1`, so I think it would still be a good idea to have a fix for that case in boot2docker as well.

tianon:
Yeah, I've got a script that I think is reasonable for fixing this in b2d regardless of whether Machine adds a serial port -- I was going to work on it more today and then we had some Alpine-bump-related explosions in official images that took over my attention.  I'll make sure to get the PR open before the next RC and poke all you fine folks here to review it. :smile: :innocent:

PR #1227, opened 2016-12-13
Switch from tinycorelinux.net to ibiblio.org (tinycorelinux.net is currently down)

tianon:


tianon:
Got this address from http://wiki.tinycorelinux.net/wiki:mirrors, which is up. :+1:

PR #1232, opened 2017-01-05
Adjust quoting/whitespace/etc of a few Dockerfile bits

tianon:


Dockerfile
```diff
@@ -272,63 +274,52 @@ RUN chroot "$ROOTFS" prltoolsd -V
...
-# Change MOTD
-RUN mv $ROOTFS/usr/local/etc/motd $ROOTFS/etc/motd
-
-# Make sure we have the correct bootsync
-RUN mv $ROOTFS/boot*.sh $ROOTFS/opt/ && \
-	chmod +x $ROOTFS/opt/*.sh
-
-# Make sure we have the correct shutdown
-RUN mv $ROOTFS/shutdown.sh $ROOTFS/opt/shutdown.sh && \
-	chmod +x $ROOTFS/opt/shutdown.sh
```
tianon: All these `mv` bits were happening after we `COPY rootfs/rootfs ...` which happens after we extract upstream's rootfs already, so our `COPY` already overlays appropriately and we don't need to move things after the fact. :smile:

tianon: (except `dhcp.sh`, which comes from the TCL rootfs and we move it into `/etc/rc.d` and invoke it directly ourselves)

tianon:
Travis is happy :smile:

PR #1233, opened 2017-01-05
Update aufs-util to an appropriately newer version (and compilation to use the proper build flags)

tianon:


tianon:
Yay, success! :+1:

PR #1234, opened 2017-01-05
Add new "forgiving-getty" wrapper which verifies that a given TTY is actually a TTY before spawning getty on it (repeatedly)

tianon:
Closes #1221
Closes: docker/machine#3711

cc @nathanleclaire @jonathanperret @simonvanderveldt

### DIFF
--- a/rootfs/rootfs/usr/local/bin/autologin
+++ b/rootfs/rootfs/usr/local/bin/autologin
@@ -1,0 +1,4 @@
+#!/bin/sh
+set -e
+
+exec /bin/login -f docker
--- a/rootfs/rootfs/usr/local/bin/forgiving-getty
+++ b/rootfs/rootfs/usr/local/bin/forgiving-getty
@@ -19,12 +19,18 @@ fi
 # https://github.com/systemd/systemd/blob/6a47fd894d601f7e8e88dec4cb35dfb7d7c15eff/src/getty-generator/getty-generator.c#L94-L116
 verify_tty() {
 	local dev="/dev/$1"
+	# can we read the given TTY file at all?
 	test -r "$dev" || return 1
+	# we can! let's open it to file descriptor 42
+	# (random file descriptor not likely to be in use already)
 	exec 42<"$dev" || return 1
 	local ret=1
+	# now that we've got it open, let's use "test" to check whether it's a TTY
+	# (we can't test a file directly, which is why we opened it first)
 	if test -t 42; then
 		ret=0
 	fi
+	# finally, close our file descriptor (no longer necessary to keep open)
 	exec 42<&- || return 1
 	return "$ret"
 }
--- a/rootfs/rootfs/usr/local/bin/forgiving-getty
+++ b/rootfs/rootfs/usr/local/bin/forgiving-getty
@@ -1,0 +1,38 @@
+#!/bin/sh
+set -e
+
+usage() {
+	self="$(basename "$0")"
+	cat <<-EOUSAGE
+		usage: $self <tty>
+		   ie: $self ttyS0
+		       $self ttyS1
+	EOUSAGE
+}
+
+tty="$1"
+if ! shift || [ -z "$tty" ]; then
+	usage >&2
+	exit 1
+fi
+
+# https://github.com/systemd/systemd/blob/6a47fd894d601f7e8e88dec4cb35dfb7d7c15eff/src/getty-generator/getty-generator.c#L94-L116
+verify_tty() {
+	local dev="/dev/$1"
+	test -r "$dev" || return 1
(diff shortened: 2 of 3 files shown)


rootfs/rootfs/usr/local/bin/forgiving-getty
```diff
@@ -0,0 +1,38 @@
...
+}
+
+tty="$1"
+if ! shift || [ -z "$tty" ]; then
+	usage >&2
+	exit 1
+fi
+
+# https://github.com/systemd/systemd/blob/6a47fd894d601f7e8e88dec4cb35dfb7d7c15eff/src/getty-generator/getty-generator.c#L94-L116
+verify_tty() {
```
nathanleclaire: I feel like this could use at least a bit more explanation (don't need to go overboard since it doesn't get touched too much) -- I'm not sure what 42 refers to below for instance

nathanleclaire: Other than the secret of the universe, obviously.

tianon: Good point.  How about something like this:
```diff
diff --git a/rootfs/rootfs/usr/local/bin/forgiving-getty b/rootfs/rootfs/usr/local/bin/forgiving-getty
index 8dacac6..4523558 100755
--- a/rootfs/rootfs/usr/local/bin/forgiving-getty
+++ b/rootfs/rootfs/usr/local/bin/forgiving-getty
@@ -19,12 +19,18 @@ fi
 # https://github.com/systemd/systemd/blob/6a47fd894d601f7e8e88dec4cb35dfb7d7c15eff/src/getty-generator/getty-generator.c#L94-L116
 verify_tty() {
 	local dev="/dev/$1"
+	# can we read the given TTY file at all?
 	test -r "$dev" || return 1
+	# we can! let's open it to file descriptor 42
+	# (random file descriptor not likely to be in use already)
 	exec 42<"$dev" || return 1
 	local ret=1
+	# now that we've got it open, let's use "test" to check whether it's a TTY
+	# (we can't test a file directly, which is why we opened it first)
 	if test -t 42; then
 		ret=0
 	fi
+	# finally, close our file descriptor (no longer necessary to keep open)
 	exec 42<&- || return 1
 	return "$ret"
 }
```

nathanleclaire:
LGTM

image

tianon:
hahahahaha :heart:

(Merging since I've tested this successfully and need to go build `rc5`, and want this included :+1:)

rootfs/rootfs/usr/local/bin/autologin
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
simonvanderveldt: @tianon did you add this file on purpose?
I already have one with boot2docker 1.12.5 which contains the following
```
$ cat /usr/local/bin/autologin
#!/bin/sh
/bin/login -f docker
```

rootfs/rootfs/usr/local/bin/forgiving-getty
simonvanderveldt: Nice! Good to have an explanation, was pretty much magic without it ;)

jonathanperret:
:+1:

rootfs/rootfs/usr/local/bin/autologin
tianon: Yeah, if you look at the `Dockerfile` changes in this PR, it used to be written via `echo` in the `Dockerfile` itself, and I simply moved it to an explicit file in the repo and added `exec` in front of that `/bin/login` so that the extra `/bin/sh` process doesn't stick around for every `getty`. :+1:

tianon:
```diff
 # Add serial console
 -RUN echo "#!/bin/sh" > $ROOTFS/usr/local/bin/autologin && \
 -	echo "/bin/login -f docker" >> $ROOTFS/usr/local/bin/autologin && \
 -	chmod 755 $ROOTFS/usr/local/bin/autologin && \
 -	echo 'ttyS0:2345:respawn:/sbin/getty -l /usr/local/bin/autologin 9600 ttyS0 vt100' >> $ROOTFS/etc/inittab && \
 -	echo 'ttyS1:2345:respawn:/sbin/getty -l /usr/local/bin/autologin 9600 ttyS1 vt100' >> $ROOTFS/etc/inittab
```

simonvanderveldt: Sorry, totally missed that you just moved it. It was a long week ;)

tianon: No worries -- can definitely relate :sweat_smile:

PR #1249, opened 2017-03-28
Only attempt to install cert files if they exist

whitlaaa:
While `install-ca-certs` allows for either `pem` or `crt` files, if no `pem` file is found it will attempt to install `*.pem` which more than likely doesn't exist. This causes a failure when installing a new `crt` file (i.,e., `corporate.crt`) when there are no `pem` files found.

This change simply checks whether `$cert_file` exists and if not, it continues to the next loop iteration.


tianon:
Simple fix, I like it! :+1:

Thanks! :smile:

PR #1255, opened 2017-04-24
change kernel_config for iotop with I/O accounting support

qxo:
fixed for iotop: add CONFIG_TASK_XACCT=y and  CONFIG_TASK_IO_ACCOUNTING=y in kernel_config
Otherwise iotop fail with:
```
Could not run iotop as some of the requirements are not met:
- Linux >= 2.6.20 with
  - I/O accounting support (CONFIG_TASKSTATS, CONFIG_TASK_DELAY_ACCT, CONFIG_TASK_IO_ACCOUNTING)
```


kernel_config
```diff
@@ -121,6 +121,8 @@ CONFIG_BSD_PROCESS_ACCT_V3=y
 CONFIG_TASKSTATS=y
 CONFIG_TASK_DELAY_ACCT=y
 # CONFIG_TASK_XACCT is not set
```
yosifkit: This line should be deleted, since it is no longer applicable.

qxo: ok

tianon:
Looks good, thanks! (and matches Debian's configuration for these flags)

PR #1271, opened 2017-08-22
Enable CAN drivers in kernel_config

carlosgalvezp:
Required for applications that use CAN communication.
Successfully tested with `can-utils` and CAN hardware.


kernel_config
```diff
@@ -1293,10 +1293,17 @@ CONFIG_BAYCOM_SER_FDX=m
...
-# CONFIG_CAN is not set
 CONFIG_IRDA=m
 
 #
+# CAN drivers
+#
+CONFIG_CAN=y
+CONFIG_CAN_RAW=y
+CONFIG_CAN_BCM=y
+CONFIG_CAN_VCAN=y
```
tianon: I'm OK with the general idea, but I think we should do `=m` instead of `=y`, similar to Debian: https://sources.debian.net/src/linux/stretch/debian/config/config/#L6388-L6394

```
##
## file: net/can/Kconfig
##
CONFIG_CAN=m
CONFIG_CAN_RAW=m
CONFIG_CAN_BCM=m
CONFIG_CAN_GW=m
```

Can you think of a reason why they don't set `CONFIG_CAN_VCAN`, and instead set `CONFIG_CAN_GW`?  Any reason we shouldn't simply match what they're doing?  (Debian's certainly got more manpower behind their kernel maintenance than we do in b2d)

carlosgalvezp: Great!

I agree, the Debian configuration looks good and works for us. As you mention, they probably have good reasons to do it that way.

Regarding `CONFIG_CAN_VCAN` vs `CONFIG_CAN_GW`, my guess is that the target application is more likely to use real CAN hardware instead of a virtualized one. For our application we don't really need VCAN since we use the hardware, so I guess we could leave it disabled until someone else needs it.

I'll update the pull request with the Debian configuration.

tianon:
LGTM!  Thanks!

carlosgalvezp:
Thank you!

Issue #1276, opened 2017-09-30
automated_script


Raboo:
Hi,

I just followed the documentation and setup ipxe to boot boot2docker with an automated script.
After little debugging I see that it's been disabled since 3 years ago.
https://github.com/boot2docker/boot2docker/blob/master/rootfs/rootfs/opt/bootscript.sh#L84

Is it possible to rewrite this to a /bin/sh script or should the documentation mentioning the possibility of an automated script be changed?

tianon:
IMO we should change the docs -- boot2docker is in pretty much just maintenance mode.  We're happy to accept bugfixes and simple PRs (especially related to the primary usage with `docker-machine`), but any additional functionality outside that is probably out-of-scope (and we'd especially like to keep the already-fragile boot process as simple as possible).

Raboo:
Ok, but I've already fixed the automated_script function via PR #1277

PR #1277, opened 2017-09-30
convert automated_script.sh to shell script

Raboo:
This solves #1276.
I've rewritten automated_script.sh to work in a non-bash shell.


tianon:
Duplicate of https://github.com/boot2docker/boot2docker/pull/1047.

PR #1280, opened 2017-10-20
MIPv6 Support

vk496:
Hello,

To run nodes with IPv6 Mobility inside Docker, is necessary that the host kernel have some kernel flags enabled.

With this modules, we can run **mipv6d** inside containers without problems.

The list of modules activated is documented here: [http://git.umip.org/umip/umip.git/plain/INSTALL.kernel](http://git.umip.org/umip/umip.git/plain/INSTALL.kernel)

I tested and works great.

PD: I don't know if `CONFIG_EXPERIMENTAL=y` is deprecated or not, because it not appeared in the built kernel.

Salu2


kernel_config
```diff
@@ -791,26 +791,26 @@ CONFIG_INET_UDP_DIAG=m
...
 CONFIG_DEFAULT_TCP_CONG="cubic"
 # CONFIG_TCP_MD5SIG is not set
-CONFIG_IPV6=m
+CONFIG_IPV6=y
 CONFIG_IPV6_ROUTER_PREF=y
 # CONFIG_IPV6_ROUTE_INFO is not set
 # CONFIG_IPV6_OPTIMISTIC_DAD is not set
 CONFIG_INET6_AH=m
-CONFIG_INET6_ESP=m
+CONFIG_INET6_ESP=y
```
tianon: Why is this being changed from `=m`?  Debian uses `=m` here, so I'd rather stick to `=m` everywhere we can (especially when Debian's kernel does).

kernel_config
```diff
@@ -791,26 +791,26 @@ CONFIG_INET_UDP_DIAG=m
...
-CONFIG_IPV6_MIP6=m
+CONFIG_IPV6_MIP6=y
 # CONFIG_IPV6_ILA is not set
 CONFIG_INET6_XFRM_TUNNEL=m
 CONFIG_INET6_TUNNEL=m
 CONFIG_INET6_XFRM_MODE_TRANSPORT=m
 CONFIG_INET6_XFRM_MODE_TUNNEL=m
 CONFIG_INET6_XFRM_MODE_BEET=m
-CONFIG_INET6_XFRM_MODE_ROUTEOPTIMIZATION=m
+CONFIG_INET6_XFRM_MODE_ROUTEOPTIMIZATION=y
```
tianon: Same story on this one -- Debian uses `=m` so I think we ought to also unless we've got a good reason not to.

kernel_config
```diff
@@ -791,26 +791,26 @@ CONFIG_INET_UDP_DIAG=m
...
 # CONFIG_IPV6_VTI is not set
 CONFIG_IPV6_SIT=m
 # CONFIG_IPV6_SIT_6RD is not set
 CONFIG_IPV6_NDISC_NODETYPE=y
-CONFIG_IPV6_TUNNEL=m
+CONFIG_IPV6_TUNNEL=y
```
tianon: Same story again; Debian uses `=m` here.

kernel_config
```diff
@@ -4822,3 +4822,7 @@ CONFIG_FONT_8x16=y
 CONFIG_ARCH_HAS_SG_CHAIN=y
 CONFIG_ARCH_HAS_PMEM_API=y
 CONFIG_ARCH_HAS_MMIO_FLUSH=y
+
+CONFIG_EXPERIMENTAL=y
+CONFIG_IPV6_ADVANCED_ROUTER=y
+CONFIG_ARPD=y
```
tianon: Do these even exist for our version of the kernel?  I don't see any of them in Debian's 4.9 config (but we use 4.4).

vk496:
Hello.

I rebuild the config more clean. I downloaded the proper version and use `menuconfig` instead of direct changes.

Tested and still working the MIPv6.

Salu2

tianon:
Looks good, thanks! :+1:

PR #1288, opened 2017-12-04
fixed kernel version

speculatrix:
the kernel you now get is much newer

### DIFF
--- a/rootfs/rootfs/etc/rc.d/automount
+++ b/rootfs/rootfs/etc/rc.d/automount
@@ -10,14 +10,15 @@ echo $BOOT2DOCKER_DATA
 if [ ! -n "$BOOT2DOCKER_DATA" ]; then
     echo "Is the disk unpartitioned?, test for the 'boot2docker format-me' string"
     
+    # FIXME! this doesn't handle multiple disks
+
     # Is the disk unpartitioned?, test for the 'boot2docker format-me' string
     UNPARTITIONED_HD=`fdisk -l | grep "doesn't contain a valid partition table" | head -n 1 | sed 's/Disk \(.*\) doesn.*/\1/'`
 
     if [ -n "$UNPARTITIONED_HD" ]; then
         # Test for our magic string (it means that the disk was made by ./boot2docker init)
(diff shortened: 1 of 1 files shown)


doc/BUILD.md
```diff
@@ -4,7 +4,7 @@ How to build boot2docker.iso locally
 The boot2docker.iso is built with Docker, via a Dockerfile.
 
 During `docker build` we
-* fetch, patch with AUFS support and build the 3.15.3 Linux Kernel with Tiny Core base configuration
+* fetch, patch with AUFS support and build the 4.4.100 Linux Kernel with Tiny Core base configuration
```
tianon: Ouch, nice catch!  Since we update the kernel relatively frequently and I usually stay on top of updating `README.md`, I think I'd rather adjust this to drop the version number, and interested folks can dig into either the `Dockerfile` or the `README` to find out the specific version number.

rootfs/rootfs/etc/rc.d/automount
```diff
@@ -10,14 +10,15 @@ echo $BOOT2DOCKER_DATA
...
     # Is the disk unpartitioned?, test for the 'boot2docker format-me' string
     UNPARTITIONED_HD=`fdisk -l | grep "doesn't contain a valid partition table" | head -n 1 | sed 's/Disk \(.*\) doesn.*/\1/'`
 
     if [ -n "$UNPARTITIONED_HD" ]; then
         # Test for our magic string (it means that the disk was made by ./boot2docker init)
-        HEADER=`dd if=$UNPARTITIONED_HD bs=1 count=${#MAGIC} 2>/dev/null`
-
-        if [ "$HEADER" = "$MAGIC" ]; then
+        echo "$MAGIC" | cmp -s -n ${#MAGIC} - $UNPARTITIONED_HD
+        if [ $? -eq 0 ]; then
```
tianon: Looks like an unrelated change snuck into this PR. :smile:

tianon:
Fixed as part of https://github.com/boot2docker/boot2docker/pull/1332. :+1:

PR #1292, opened 2017-12-10
Update from jessie to stretch

tianon:


tianon:
This saves ~100MB of image size (by switching to `-slim`).

PR #1294, opened 2017-12-18
Fix fsync on dirs in Parallels Shared Folders

romankulikov:
Previously prl_fs didn't provide fsync handler for directory operations
(but for files did). As a result respective syscalls failed. That
caused, for example, problems with running PostgreSQL in containers over
database on a shared volume.

This patch is really a kludge to workaround the issue. And should be
removed when Parallels Tools with fixed prl_fs is incorporated.

Parallels/docker-machine-parallels#71

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -281,6 +281,11 @@ RUN set -ex \
 	&& cd /prl_tools \
 	&& cp -Rv tools/* $ROOTFS \
 	\
+	# Kludge to fix fsync on directory in prl_fs:
+	#   https://github.com/Parallels/docker-machine-parallels/issues/71
+	# Should be removed when fixed in official Parallels Tools.
+	&& sed -E -i 's/(^\t.llseek\t\t= generic_file_llseek,$)/\1\n\t.fsync = noop_fsync,/' \
+		kmods/prl_fs/SharedFolders/Guest/Linux/prl_fs/file.c \
 	&& KERNEL_DIR=/linux-kernel/ KVER="$KERNEL_VERSION" SRC=/linux-kernel/ PRL_FREEZE_SKIP=1 \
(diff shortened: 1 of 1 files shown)


tianon:
Any idea what the timeframe looks like on an upstream fix?  This seems OK in the interim, but we haven't had a Parallels Tools update in a long time (and I can't seem to find any reference online which would help us update to a newer version even if one were available, given that directory browsing on download.parallels.com is disabled), so I'm just concerned we'll be carrying this kludge for a long time. :disappointed:

romankulikov:
> Any idea what the timeframe looks like on an upstream fix?

Sorry, no official timeframes from Parallels.

> This seems OK in the interim, but we haven't had a Parallels Tools update in a long time (and I can't seem to find any reference online which would help us update to a newer version even if one were available, given that directory browsing on download.parallels.com is disabled), so I'm just concerned we'll be carrying this kludge for a long time

I want to note here that from version 12.1.3 (which is used currently in boot2docker) to the latest 13.2.0 (to the moment) there were no any significant improvements on the Parallels Tools side for boot2docker. The most changes were done on the host side of Parallels Desktop. So updating Parallels Tools wouldn't get any real benefit. Much better for the end users to upgrade their Parallels Desktop instances on their Macs.

But sure I'm going provide pull request with the version of Parallels Tools which will contain fix of fsync issue and get rid of the kludge.

romankulikov:
Can someone point why did build check fail? I've tried to reproduce it several times in different environments but with no luck – all builds are ok.

tianon:
The Travis failure is harmless -- if it builds successfully locally, you're good.  The build takes so long that often Travis will fail simply due to that. :smile:

Thanks for the additional information. :heart:

This looks good! :+1:

PR #1298, opened 2018-01-16
Remove -D default debug option

### LINKED ISSUE #1297: Remove debug flag from default daemon options
Debug mode (`-D`) is set on by default for the daemon in boot2docker, although this prevents people from editing it through the config.json if they wish (as daemon flags cannot be overridden by config flags). Because `docker-machine` provides the ability to pass daemon flags then users can easily add this functionality if they want, but this also allows the freedom to use config.json as well.

TL;DR removing the `-D` default gives users greater flexibility.

Thoughts?

johnharris85:
Removes the `-D` flag from `dockerd` so that users can optionally add this either via `daemon.json` or via `--engine-opt`.

Closes #1297


tianon:
Awesome, thanks!! :+1:

PR #1307, opened 2018-03-01
Enable SMB2 kernel support

lmakarov:
This allows mounting Windows SMB shares with `vers=2.1` argument, otherwise only `vers=1.0` works.

SMB1.0 support has been disabled by default with Windows 10 Fall Creators Update 1709:
https://support.microsoft.com/en-us/help/4034314/smbv1-is-not-installed-windows-10-and-windows-server-version-1709

See https://github.com/blinkreaction/boot2docker-vagrant/issues/66#issuecomment-368459166 for more details.


lmakarov:
I rebased the PR today. Wonder if it can be accepted. It looks to be a trivial change.

ping @tianon

tianon:
Sorry for the delay, this seems reasonable :+1: 

For completeness, here's Debian's value: (https://sources.debian.org/src/linux/4.9.82-1+deb9u3/debian/config/config/#L5377)

```
CONFIG_CIFS_SMB2=y
```

PR #1317, opened 2018-04-09
Enable CONFIG_CEPH_FS=m

### LINKED ISSUE #1254: Include CEPH kernel support
It seems that CEPH filesystem is not supported from boot2docker kernel module:
`# CONFIG_CEPH_FS is not set`

It would be possible to include this? Ceph backend is quite common and it could be useful to mount a shared file system from the docker service.

tianon:
Closes #1254


tianon:
Ok, built this successfully, and as usual for these types of PRs (for unknown reasons), final ISO size is 100% unchanged. :+1:

PR #1318, opened 2018-04-16
Add Transparent Proxy

MarcinLachowicz:
Running Docker in the environment with HTTP(S) proxy, which is still common in the corporate networks,
is hard because you have to provide the proxy address and authentication in multiple places.

Internet access is needed for example for:

* pulling images from Docker Hub
* downloading packages while building Docker images
* containers communication to the Internet

Setting `http_proxy` and `https_proxy` environment variables is usually not sufficient - not all programs support it.
To address this problem Boot2docker is capable of redirecting HTTP and HTTPS traffic to the proxy transparently,
but it requires some additional Docker Machine setup.

In the nutshell, either provide HTTP(S) proxy url via `--engine-env TRANSPARENT_HTTP_PROXY=user:password@proxy-host.com`
when creating a new Docker Machine or in `~/.docker/machine/machines/{machineName}/config.json` for existing machines.

This solution uses `iptables` and [redsocks](https://github.com/darkk/redsocks) to redirect tcp traffic
to [cntlm](http://cntlm.sourceforge.net).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,7 @@ RUN set -eux; \
 		golang \
 		isolinux \
 		kmod \
+		libssl1.0-dev libevent-dev=2.0.21-stable-3 \
 		p7zip-full \
 		pkg-config \
 		squashfs-tools \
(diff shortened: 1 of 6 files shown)


tianon:
I think this is really neat, but unfortunately the added complexity is not something we're willing to support in boot2docker natively.

I'd recommend that folks who need this either build a custom ISO (which isn't likely to be well-supported) or better yet either update their host system to do the transparent proxying or base their deployments on a distribution which handles customization as a first-class citizen such as Debian or Ubuntu.

MarcinLachowicz:
Hello,
I agree with you, it is more like bypassing the problem instead of solving it e.g. by using Debian or Ubuntu. Unfortunately, some people still have to use `boot2docker` and corporate proxy. That is why we created such feature. I will keep it in our fork. 
Best regards!

PR #1324, opened 2018-07-05
Update open-vm-tools to 10.2.5-8068406

masaeedu:
The currently bundled version of open-vm-tools has an issue where only 16 IPs are going to be reported to the vmware hypervisor, and the link local IP addresses of veth* NICs are included. The end result is that if you start up a large compose file, the VM's own IP will be pushed out of the reported guest IPs list, and commands such as `docker-machine env` and `docker-machine ip` will hang.

See vmware/open-vm-tools#100


tianon:
LGTM, thanks!

PR #1332, opened 2018-09-10
Add new ISO generation method intended to replace existing Dockerfile

tianon:
This is a whole new ISO that's intended to eventually replace the existing one (and should explicitly be compatible completely -- it's just a cleaner way to build TCL into an ISO via Docker).

The main gist of these changes is using TCL's own `tce-load` with a minor patch in order to install packages in our target environment, which allows us to largely ignore their dependencies (ie, the entire purpose of a package manager).

This also includes a bump to Linux 4.14.69 and the requisite addition of `haveged` which is noted/discussed over in https://github.com/boot2docker/boot2docker/pull/1322.


tianon:
There's an example build up at https://gist.github.com/tianon/1ff294a629942e6f0a019c532b3b2734/raw/97e8a794d019ae77b1c9636d789f790740b4ed1b/boot2docker.iso.

@legal90 any chance you could test this via Parallels and ensure it does the right thing? :pray: :heart:

(Also, is there somewhere public I could link to for Parallels Tools releases in https://github.com/tianon/boot2docker-legacy/blob/b91259e06cc066d27b3931585465b161a26c845d/tcl/files/make-b2d-iso.sh#L36?)

tianon:
So far, the best I can find is something like https://github.com/Homebrew/homebrew-cask-versions/blob/master/Casks/parallels13.rb, which at least would give us a source for updating this version number, but it isn't official (and I'm not sure where they're getting their version numbers from -- probably just stealing it from the application itself after it updates).

tianon:
> Hm.. It seems that boot2docker.iso image which you mentioned launches docker daemon with listening to UNIX socket only. It doesn't listen to TCP socket, so I can't verify it with docker-machine 0.15.0.

Whoops, that was an oversight, sorry!  I've updated the branch and the gist with a newer ISO which should fix that: https://gist.github.com/tianon/1ff294a629942e6f0a019c532b3b2734 (https://gist.github.com/tianon/1ff294a629942e6f0a019c532b3b2734/raw/f84b9fbb224dbe2fbdf6d7c8a87d447d2ec4a961/boot2docker.iso)

I've successfully tested this one with Hyper-V via `docker-machine`, so fingers crossed it's all fixed. :sweat_smile:


> Currently the version numbers which we use here in Dockerfile (...) are updated on-demand, e.q. the new tar is published only when it's decided to promote it to boot2docker.

Ah, fair enough!  I won't worry about it then. :innocent: :heart:

legal90:
@tianon Thank you! I tested the new iso and it works like a charm under the latest Parallels Desktop(v14.0.0 build 45124 at this moment) ! 🎉
There is no delay of `dockerd` start anymore.

tianon:
Excellent, thanks for testing and confirming, @legal90! :+1:

tianon:
I was thinking of publishing this alongside the existing `boot2docker.iso` for a couple releases, but I'm currently leaning more towards replacing the existing `boot2docker.iso` entirely and just going all-in here.

tianon:
As another point in favor of this, the resulting ISO not only has more useful tools installed (like properly configured `bash-completion`), but is a few MB smaller. :+1:

tianon:
Yep, see https://github.com/boot2docker/boot2docker/pull/1350 and
https://github.com/boot2docker/boot2docker/pull/1370, which should restore
proper VMware support in the next build.

Issue #1326, opened 2018-08-02
Deprecate AUFS?


tianon:
The latest release of Docker for Windows ("Docker Community Edition 18.06.0-ce-win70 2018-07-25") has officially deprecated AUFS with the following notice (https://docs.docker.com/docker-for-windows/release-notes/#docker-community-edition-18060-ce-win70-2018-07-25):

> AUFS storage driver is deprecated in Docker Desktop and AUFS support will be removed in the next major release. You can continue with AUFS in Docker Desktop 18.06.x, but you will need to reset disk image (in Settings > Reset menu) before updating to the next major update. You can check documentation to [save images](https://docs.docker.com/engine/reference/commandline/save/#examples) and [backup volumes](https://docs.docker.com/storage/volumes/#backup-restore-or-migrate-data-volumes)

I'm thinking it's probably sane for us to follow suit and have 18.06.x be the final release that supports AUFS (especially since it repeatedly makes kernel upgrades / choosing a kernel version challenging).

tianon:
Ok, let's do this.  Release notes for 18.06.x officially updated: https://github.com/boot2docker/boot2docker/releases/tag/v18.06.0-ce

I'm going to leave this issue open for a bit as a catch-all to hopefully provide slightly better visibility for this change.

kaosagnt:
So far changing /var/lib/boot2docker/profile to overlay2

No issues loading an image and attaching a container.

Question. How does /var/lib/boot2docker/profile get created in a newly created vm. Does that come from docker-machine?

docker@kernel-414:~$ docker info
Containers: 1
 Running: 0
 Paused: 0
 Stopped: 1
Images: 1
Server Version: 18.06.0-ce
Storage Driver: overlay2
 Backing Filesystem: extfs
 Supports d_type: true
 Native Overlay Diff: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host macvlan null overlay
 Log: awslogs fluentd gcplogs gelf journald json-file logentries splunk syslog
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: d64c661f1d51c48782c9cec8fda7604785f93587
runc version: 69663f0bd4b60df09991c08812a60108003fa340
init version: fec3683
Security Options:
 seccomp
  Profile: default
Kernel Version: 4.14.60-boot2docker
[...]

PR #1333, opened 2018-09-11
Explicitly ignore requests for AUFS (since they'll just fail to start dockerd anyhow)

tianon:
Closes https://github.com/boot2docker/boot2docker/issues/1326


tianon:
This makes the `docker-machine`, etc updates much less urgent. :+1:

PR #1343, opened 2018-11-06
Fix vboxsf.ko inclusion (fixes VirtualBox shared folders)

### LINKED ISSUE #1342: boot2docker v18.09.0-rc1 VirtualBox v5.2.20 share directory error
VirtualBox v5.2.20 Unable to load shared directory。

sudo -i mount -t vboxsf E_DRIVE /mnt/sda1/share
/sbin/mount.vboxsf: mounting failed with the error: No such device

tianon:
Fixes #1342

Thanks for testing and reporting, @luo12yan! :heart:


tianon:
Looks like we also need more handling around `resolv.conf`, since sometimes it ends up empty. :confused:

PR #1348, opened 2018-11-09
Fix volume ID for docker-machine

### LINKED ISSUE #1347: v18.09.0 does not embed version number in ISO header
docker-machine looks for a version number in the boot2docker ISO header. See here:
https://github.com/docker/machine/blob/master/libmachine/mcnutils/b2d.go#L310

v18.09.0 does not appear to be embedding a version number in the same way as previously.
Offset 0x8028 in v18.09.0 ISO has the string `Boot2Docker`
Offset 0x8028 in v18.06.1-ce ISO has the string `b2d-v18.06.1-ce`

The consequence of this is that docker-machine repeatedly logs
`Unable to get the local Boot2Docker ISO version:  Did not find prefix \"-v\" in version string"`
`Default Boot2Docker ISO is out-of-date, downloading the latest release...`

And, more seriously, it does re-download the boot2docker ISO for every machine created!

tianon:
Fixes #1347

```console
$ isoinfo -d -i boot2docker.iso
CD-ROM is in ISO 9660 format
System id: 
Volume id: b2d-v18.09.0
Volume set id: 
Publisher id: 
Data preparer id: XORRISO-1.4.6 2016.09.16.133001, LIBISOBURN-1.4.6, LIBISOFS-1.4.6, LIBBURN-1.4.6
Application id: BOOT2DOCKER
Copyright File id: 
Abstract File id: 
Bibliographic File id: 
Volume set size is: 1
Volume set sequence number is: 1
Logical block size is: 2048
Volume size is: 23040
El Torito VD version 1 found, boot catalog is in sector 49
NO Joliet present
Rock Ridge signatures version 1 found
Eltorito validation header:
    Hid 1
    Arch 0 (x86)
    ID ''
    Key 55 AA
    Eltorito defaultboot header:
        Bootid 88 (bootable)
        Boot media 0 (No Emulation Boot)
        Load segment 0
        Sys type 0
        Nsect 4
        Bootoff 32 50
```

### DIFF
--- a/files/make-b2d-iso.sh
+++ b/files/make-b2d-iso.sh
@@ -6,9 +6,12 @@ find -not -name '*.tcz' \
 	| xz -9 --format=lzma --verbose --verbose --threads=0 --extreme \
 	> /tmp/iso/boot/initrd.img
 
+# volume label (https://github.com/boot2docker/boot2docker/issues/1347)
+volumeLabel="b2d-v$DOCKER_VERSION"
+
 xorriso \
 	-as mkisofs -o /tmp/boot2docker.iso \
-	-A 'Boot2Docker' -V 'Boot2Docker' \
+	-A 'Boot2Docker' -V "$volumeLabel" \
 	-isohybrid-mbr /tmp/isohdpfx.bin \
 	-partition_offset 16 \
 	-b isolinux/isolinux.bin \


tianon:
See https://github.com/docker/machine/blob/821e8c76f048ca9323ecc9440c0e3d3d5ec4906d/libmachine/mcnutils/b2d.go#L291-L321 for the exact bit of code in `docker-machine` that checks for this.

pavanssetty:
Why can't the ISO file be fixed too? I am not sure how I can get this fix on my computer.

tianon:
It will be in the next release of boot2docker.

vchebotarev:
@pavanssetty https://github.com/boot2docker/boot2docker/issues/1347#issuecomment-447812960
this is the only way

Issue #1366, opened 2018-11-29
Long boot time on VirtualBox with Docker-Toolbox


TeeJayD:
I'm having issues with long boot times while using the latest boot2docker version with Docker-Toolbox.
It is a corporate environment where it blocks dns queries besides our own local dns server, and i've noticed that the ntp server always sends dns queries towards 8.8.8.8, 1.1.1.1 and 1.0.0.1, stalling the boot for at least two minutes. It takes so much time that the "Waiting for IP" step fails.
Is there any way of changing this behaviour? I've tried creating a VM with --engine-opts dns=OUR_DNS and it didn't work.

Verifying the traffic i've noticed it ignores the DNS setting and keeps trying to resolve through Google/Cloudflare.
```
192.168.40.17	1.0.0.1	DNS	86	Standard query 0xc0a0 A 0.boot2docker.pool.ntp.org
192.168.40.17	1.0.0.1	DNS	86	Standard query 0xc0a0 A 0.boot2docker.pool.ntp.org
192.168.40.17	8.8.8.8	DNS	86	Standard query 0xc0a0 A 0.boot2docker.pool.ntp.org
192.168.40.17	8.8.8.8	DNS	86	Standard query 0xc0a0 A 0.boot2docker.pool.ntp.org
```
Edit:
With some further testing, i've noticed it only happens during boot. If i try to do any dns queries using nslookup after it finally boots, it uses the correct DNS server.

tianon:
See https://github.com/boot2docker/boot2docker/pull/1367, which should help with this somewhat.

TeeJayD:
The latest beta with #1367 applied fixes the issue.

yosifkit:
@TeeJayD, Thanks for confirming!

Issue #1330, opened 2018-08-28
unable to set NTP-servers during boot


umlaeute:
i noticed that the `ntpd` process gets started with a [hard-coded list of ntp-servers](https://github.com/boot2docker/boot2docker/blob/master/rootfs/rootfs/etc/rc.d/ntpd#L3):

    : ${NTP_SERVER:='0.boot2docker.pool.ntp.org 1.boot2docker.pool.ntp.org 2.boot2docker.pool.ntp.org 3.boot2docker.pool.ntp.org'}

nowadays, many companies block outgoing NTP-traffic (to prevent DDoS attacks), which makes this sub-optimal (eventually resulting in hangs until the ntpd times out).

i would have expected to be able to override the `NTP_SERVER` variable, but changes to `/var/lib/boot2docker/profile` seem to have no effect (that files seems to be only used by the `docker` process).

obviously, the `NTP_SERVER` value is meant to be settable (and the list of ntp-servers is only a fallback), but i don't see a way to inject any values.

kaosagnt:
Maybe something like this? This is just off the top of my head, not tested.

create /var/lib/boot2docker/bootlocal.sh

```
#!/bin/sh

pkill ntpd

NTP_SERVER=someserver /etc/rc.d/ntpd &

```


/var/lib/boot2docker/bootlocal.sh is run if it exists at boot time after ntpd is run.

umlaeute:
while this is a valid workaround, i'm not really convinced: usually I want my daemons to be configured correctly *before* i start them.

my motivation to report the problem was, that `boot2docker` took slightly longish to startup and I suspected the failing NTP-servers to be the problem.
if this is true, then your fix doesn't help at all.

also, i think that `bootsync.sh` makes the process non-deterministic (e.g. see #1331)

tianon:
If your environment takes a long time to start up, you're probably blocking
ping probes (I get that very issue while testing in QEMU with user mode
networking which doesn't support ping).

umlaeute:
@tianon pings go through nicely.
also note that "slightly longish" might sound exaggerated: boot time is about *30 seconds*. (but given that this is a VM that is spawned automatically "just so" (as a GitLab-CI runner), every single second counts.

tianon:
See https://github.com/boot2docker/boot2docker/pull/1367 which might help with this.

PR #1367, opened 2018-12-04
Adjust DHCP to be slightly less asynchronous, to log to /var/log/dhcp.log, and to apply NTP servers from DHCP to our local configuration

tianon:
This might help with #1366 and should help with #1330 (by allowing DHCP values to overwrite the default provided NTP configuration).


tianon:
I've tested this successfully with both QEMU and VirtualBox (although minus the NTP bit because I don't have a good setup to test that).

At the very least, this will hopefully help with DNS from DHCP getting dropped into `/etc/resolv.conf` sooner so that `ntpd` uses it appropriately (which is the issue described at #1366).

PR #1381, opened 2019-02-21
enable tc-restore support

hiobama:


tianon:
I'm not in favor of this frankly because it's yet more behavior that I won't be testing (and there's already mountains of that type of functionality here that I'm not happy about but we can't really get away from at this point).

From what I can tell about this tiny core feature, it sounds like you're trying to use boot2docker beyond it's limited intended purpose (namely that of being a limited/lightweight distro for `docker-machine`), which is typically when I recommend folks turn to a more suitable distribution such as Debian, Ubuntu, or even Alpine.

In your case it sounds like you've got a decent amount of Tiny Core experience already, so you might even be better off using Tiny Core itself directly.

hiobama:
Alpine is great, Thank you for mentioning.

PR #1383, opened 2019-03-06
added missing mkdir for /var/lib/boot2docker/etc causing following er…

Gr33nbl00d:
…ror:

Error creating machine: Error running provisioning: ssh command error:
command : sudo /usr/bin/sethostname test && echo "test" | sudo tee /var/lib/boot2docker/etc/hostname

### DIFF
--- a/files/bootsync.sh
+++ b/files/bootsync.sh
@@ -15,6 +15,9 @@ cgroupfs-mount
 mkdir -p /var/lib/boot2docker/log
 chown docker /var/lib/boot2docker/log
 
+mkdir -p /var/lib/boot2docker/etc
+chown docker /var/lib/boot2docker/etc
+
 # wouldn't it be great if ntpd had a "log to XYZ file" option? (like crond does!)
 ntpd -d -n -q >> /var/lib/boot2docker/log/ntp.log 2>&1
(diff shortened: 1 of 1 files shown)


tianon:
Huh, I guess there is a very small race condition there between when we start OpenSSH and when we start Docker (which then creates `/var/lib/boot2docker/etc/docker`).

So, I agree with this change, but I don't think we need the `chown` line and I think we should move it up next to the bit that tests and applies the hostname in the hostname file and add a brief comment describing why we do it and a link to https://github.com/docker/machine/blob/7a9ce457496353549916e840874012a97e3d2782/libmachine/provision/boot2docker.go#L113; ala:

```bash
# make sure "/var/lib/boot2docker/etc" exists for docker-machine to write into and read "hostname" from it if it exists from a previous boot
# https://github.com/docker/machine/blob/7a9ce457496353549916e840874012a97e3d2782/libmachine/provision/boot2docker.go#L113
mkdir -p /var/lib/boot2docker/etc
if [ -s /var/lib/boot2docker/etc/hostname ]; then
	hostname="$(cat /var/lib/boot2docker/etc/hostname)"
	sethostname "$hostname"
fi
```

Gr33nbl00d:
Absolutly ok with your proposal :) I can update it on monday

PR #1390, opened 2019-05-07
Hard-coded DNS servers in udhcpc busybox script

mfranciszkiewicz:


tianon:
Can you explain the intent behind this change a bit more please?  For example, what issue did you experience, how does this fix it, and how can it be reproduced?

mfranciszkiewicz:
Sorry, this was opened here instead of the forked repository by mistake.

The issue lies with the `udhcpc` startup script. It replaces the contents of `/etc/resolv.conf` if a custom DNS server is provided with a single NS entry.

PR #1391, opened 2019-05-09
Add required kernel modules for Rancher networking

borjator:
This is the list of modules based on https://github.com/rancher/rancher/issues/10499#issuecomment-393297923 and my own tests to make Rancher networking work properly.


tianon:
Nice!

I do wonder how much of this is Rancher-specific and how much is general to Kubernetes at large (perhaps there's a better source of what's necessary for Kubernetes in general that we could glean a more complete list from).  Room for future research/improvement. :+1:

Thanks for the contribution and especially for digging up a useful source and doing testing! :+1: :heart:

PR #1402, opened 2019-10-20
Swap TCL's squashfs-tools for a static compiled version instead

tianon:
For some reason, the latest 4.4 upload of squashfs-tools in TCL 10.x fails to extract many/most .tcz files from TCL.  On the flip side, it's really easy to compile statically and the `Dockerfile` is much cleaner, so we're going with it.


tianon:
For anyone looking for more info on the failure mode, see https://github.com/boot2docker/boot2docker/commit/31e764375a64dd66a18ad3cab7aeeb35e659a532/checks?check_suite_id=272529669.

PR #1403, opened 2019-11-13
Remove haveged in favor of backported upstream kernel commit

tianon:
Quoting the `Dockerfile` comment:

> apply kernel entropy patch from 5.4; this same patch was backported in Debian in 5.3.9-1
>
>   - https://git.kernel.org/linus/50ee7529ec4500c88f8664560770a7a1b65db72b
>   - https://salsa.debian.org/kernel-team/linux/commit/c323c453b2485a33bfb33635a07f3a50bc1db1ee
>   - https://lists.debian.org/debian-boot/2019/11/msg00077.html
>
> specifically, this solves the problem of early-boot entropy (SSH key generation, for example), avoiding the need for userspace solutions like haveged

Refs https://github.com/boot2docker/boot2docker/pull/1322


yosifkit (approved):

legal90:
@tianon Do you have any plans to port it to `19.03.x` releases? Or is the next release from `master` coming soon?
It seems it could fix the issue appeared recently with the parallels driver: https://github.com/Parallels/docker-machine-parallels/issues/83
Thanks! 👍

tianon:
I wasn't planning to given that `haveged` was our solution there, but I guess it's something we could consider.  I'm wary of it simply because I haven't been able to do any heavy testing of it, so didn't want to interrupt the 19.03.x releases stability with it yet -- if you're willing to make a build and do some testing to ensure it works OK (and really does fix the problem described in that issue), I'm amenable. :+1:

legal90:
@tianon Thanks!
I have built boot2docker.iso from this PR branch (ref `ab6b28e`) and  can confirm that the issue Parallels/docker-machine-parallels#83 is fixed there. VM restarts successfully every time and `dockerd` starts very fast there. 👍

tianon:
Nice!  Backported in 8be24bc, so should be included in whatever the next release is (whether that's 19.03.x or something newer) :+1:

tianon:
Ah darn, it fails to build (should've done a build test before I committed): https://github.com/boot2docker/boot2docker/commit/8be24bcf79c8b7e1c200a86175fc8896be3256d8/checks#step:6:2862

So I guess we'd need to also update the kernel there, which might have a wider impact, but is probably worthwhile.

tianon:
Ok, resolved in cb60095dfd3e6a203289b2e65c928a21bbd0a513. :+1:

legal90:
Hi @tianon 
Does this patch have any chance to be released? Maybe at least as an RC / unstable release ?

I see that is b2d has been deprecated (https://github.com/boot2docker/boot2docker/pull/1408), but this PR and its port to the branch `19.03.x` is important for those users who still want to run Docker Machine with Parallels Desktop.

tianon:
Yeah, that's fair -- I've pushed out https://github.com/boot2docker/boot2docker/releases/tag/v19.03.12, which will be the final release and includes an updated kernel (which includes this change).

legal90:
Thank you, @tianon, for all the work you've done on boo2docker !

